PR #185, opened 2019-01-29
Read the docs

k-dovgan:
Fixes for readthedocs.org integration


setup.py
```diff
@@ -36,11 +36,17 @@ def readme():
...
+            'sphinx-argparse',
+            'sphinx_rtd_theme'
+        ],
         'development': [
             'gitpython',
             'P4',
             'sphinx',
-            'sphinx-argparse'
+            'sphinx-argparse',
+            'sphinx_rtd_theme'
```
i-keliukh: Does it generate docs with rtd theme when run locally?
Can we reference one extra from another?

k-dovgan: Yes, seems so.

PR #252, opened 2019-02-08
feat(docs): add logo and favicon

### LINKED ISSUE #238: Create logo
The logo should be in vector format to be easily scalable. We might use it as a big picture (e.g. in presentations), but we also have to put something to favicon, so we also need a pretty-looking small one.

As for the contents, it should represent the ideas of integration & connectivity, and also it'd be great if there was something like a star or stars (as it is called 'Universum' as in a reference to outer space).

k-dovgan:
Resolves #238


k-dovgan:
https://universum.readthedocs.io/en/logo/
Somehow favicon works, but logo doesn't

k-dovgan:
Also, config example taken from here: https://github.com/rtfd/sphinx_rtd_theme/blob/master/docs/conf.py

i-keliukh:
> https://universum.readthedocs.io/en/logo/
> Somehow favicon works, but logo doesn't

Logo starts to work if you remove width:auto from its style.

i-keliukh:
I believe, with width:auto it gets resized to zero.
Probably you need to set image size in the image file.

Also, please add logo of bigger size to the index.

i-keliukh:
Please, check the index page on mobile browser.
IMO, the picture is too big.

Also, please remove the link to full-size picture from index page.

k-dovgan:
https://dejavu-fonts.github.io/License.html

doc/index.rst
```diff
@@ -15,9 +15,10 @@ Project 'Universum'
     changelog_ref.rst
 
 .. image:: _static/logo.svg
-    :height: 200
-    :width: 200
+    :height: 150
+    :width: 150
     :align: right
+    :class: logo-image
```
i-keliukh: Class is not needed

Issue #257, opened 2019-02-20
Exception with backtrace from vcs modules if optional import is missing


i-keliukh:
# Description

If user attempts to use vcs, but the respective python module is not installed, the Universum correctly produces extended error, but also display big back trace.
For example, if user wants to use p4, but doesn't have perforce-p4python the back trace is displayed.
The bug appeared as a part of implementation for #131 .

----------

# Step to reproduce

* Install Universum on a clean system. Docker can be used for this
* Setup the following environment variables for p4: P4PORT, P4USER and P4PASSWD. The values do not matter.
* Initiate any operation with p4. For example, the bug can be reproduced by using submit command:
```
universum submit --project-root  "/" --commit-message "abc" --reconcile-list "test"
```

# Expected result

The Universum must display import error:
```
Error:  using VCS type 'p4' requires official Helix CLI and Pyhton package
'perforce-p4python' to be installed to the system. Please refer to `Prerequisites`
chapter of project documentation for detailed instructions
```

# Actual result

The Universum produced back trace:
```
Traceback (most recent call last):
  File "/usr/local/bin/universum", line 11, in <module>
    load_entry_point('Universum==0.17.0', 'console_scripts', 'universum')()
  File "/usr/local/lib/python2.7/site-packages/universum.py", line 80, in main
    return run(settings)
  File "/usr/local/lib/python2.7/site-packages/universum.py", line 44, in run
    main_module = construct_component(settings.main_class, settings)
  File "/usr/local/lib/python2.7/site-packages/_universum/lib/gravity.py", line 68, in construct_component
    instance.__init__(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/_universum/modules/output/output.py", line 19, in new_init
    original_init(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/_universum/modules/structure_handler.py", line 21, in new_init
    original_init(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/_universum/submit.py", line 43, in __init__
    self.vcs = self.vcs_factory()
  File "/usr/local/lib/python2.7/site-packages/_universum/lib/gravity.py", line 80, in constructor_function
    return construct_component(self.klass, instance.main_settings, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/_universum/lib/gravity.py", line 68, in construct_component
    instance.__init__(*args, **kwargs)
[...]

PR #263, opened 2019-02-22
fix(vcs): fix exit code 0 on incorrect import

k-dovgan:


i-keliukh:
This PR fixes the failure of tests introduced by a39b9c0127269689b865bc1f1850fbefc3c83750.
The fix should be considered as a part of the #257.

PR #289, opened 2019-04-04
feat(tests): reuse deployment containers in PyCharm

### LINKED ISSUE #276: Do not reset docker containers for local tests
When running CI builds, all tests should be run in separate clean docker containers, as it is implemented now. But when developing locally, all docker containers can be reused for time economy purposes (as it is done now for Perforce server container). This especially applies to deployment tests containers.

k-dovgan:
- [x] reuse 'local sources'
- [x] reuse containers
- [x] but do not reuse for deployment tests
- [x] and restart outdated containers

Resolves #276

### DIFF
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -151,7 +151,8 @@ def local_sources(tmpdir):
         source_dir = py.path.local(".work").ensure(dir=True)
     else:
         source_dir = tmpdir.mkdir("project_sources")
-    local_file = source_dir.join("readme.txt").write("This is a an empty file")
+    local_file = source_dir.join("readme.txt")
+    local_file.write("This is a an empty file")
 
     yield utils.Params(root_directory=source_dir, repo_file=local_file)
 
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -54,17 +54,20 @@ class ExecutionEnvironment(object):
 
         try:
             self._container = self._client.containers.get(self._container_id)
-            return True
+            if not utils.check_if_container_outdated(self._container):
+                return True
         except docker.errors.NotFound:
-            self._container = self._client.containers.run(self._image,
-                                                          name=self._container_id,
-                                                          command="sleep infinity",
-                                                          network_mode='host',
-                                                          volumes=self._volumes,
-                                                          environment=self._environment,
-                                                          auto_remove=True,
-                                                          detach=True)
-            return False
+            pass
+
+        self._container = self._client.containers.run(self._image,
+                                                      name=self._container_id,
+                                                      command="sleep infinity",
+                                                      network_mode='host',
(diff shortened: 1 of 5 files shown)


tests/deployment_utils.py
```diff
@@ -78,15 +95,28 @@ def assert_unsuccessful_execution(self, cmd, environment=None):
 
     def install_python_module(self, name):
         cmd = "pip --default-timeout=1200 install " + name
-        log = self.assert_successful_execution(cmd)
-        assert "Successfully installed" in log
+        log = ""
```
i-keliukh: Please make the logic of this function more straightforward

tests/deployment_utils.py
```diff
@@ -104,8 +134,24 @@ def execution_environment(request):
...
+    finally:
+        if runner:
+            runner.exit()
+
+
 @pytest.fixture()
 def local_sources(tmpdir):
+    if utils.is_pycharm():
+        tmpdir = py.path.local(".work")
+        tmpdir.remove(rec=1, ignore_errors=True)
```
i-keliukh: Why we need to remove dir?

tests/deployment_utils.py
```diff
@@ -202,14 +248,21 @@ def universum_runner(execution_environment, docker_registry_params, runner_witho
...
+    clean_execution_environment.set_image("universum_test_env", docker_registry_params)
+    runner_without_environment.set_environment(clean_execution_environment)
     yield runner_without_environment
 
 
 @pytest.fixture()
-def universum_runner_no_vcs(execution_environment, docker_registry_params, runner_without_environment):
-    execution_environment.set_image("universum_test_env_no_vcs", docker_registry_params)
-    runner_without_environment.set_environment(execution_environment)
+def universum_runner_no_p4(clean_execution_environment, docker_registry_params, runner_without_environment):
```
i-keliukh: Please use consistent naming: all similar fixtures should either use or not use the same "clean_" prefix

tests/utils.py
```diff
@@ -84,6 +89,18 @@ def get_image(request, client, params, name):
     return image
 
 
+def python_time_from_rfc3339_time(rfc3339_time):
+    return tf_from_timestamp(rfc3339_time)
+
+
+def check_if_container_outdated(container):
```
i-keliukh: Please rename this function

tests/deployment_utils.py
```diff
@@ -78,15 +95,28 @@ def assert_unsuccessful_execution(self, cmd, environment=None):
 
     def install_python_module(self, name):
```
i-keliukh: Does it make sense to always attempt to install modules, even when new container is not created?

tests/deployment_utils.py
i-keliukh: The remove fails if ".work" doesn't exist before running tests

Issue #119, opened 2019-01-29
Internal error in submit when vcs type is not set, but settings are provided for gerrit


k-dovgan:
Originally created on Tue, 28 Nov 2017 16:41:02 +0900

Steps to reproduce:
1. Set environment variables for both gerrit and p4:
   * GIT_REFSPEC
   * GIT_REPO
   * P4PASSWD
   * P4PORT
   * P4USER
2. Launch submit without setting `vcs-type`:
```
universum submit  --project-root <path> --commit-message "<msg>" --reconcile-list <full_path>
```
Expected result: error is reported regarding misconfiguration or missing parameter
Actual result: portion of internal exception is displayed (`cannot concatenate 'str' and 'NoneType' objects`)

i-keliukh:
@k-dovgan, it looks like the bug is fixed already. Please find what commit fixed the bug and close it.

i-keliukh:
My bad, the bug is not fixed. Below are logs from recent reproduce:

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/_universum/lib/utils.py", line 178, in excepted_function
    func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/_universum/lib/utils.py", line 201, in function_in_block
    return self.structure.run_in_block(func, block_name, pass_errors, self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/_universum/modules/structure_handler.py", line 131, in run_in_block
    result = operation(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/_universum/submit.py", line 52, in execute
    edit_only=self.settings.edit_only)
  File "/usr/local/lib/python2.7/dist-packages/_universum/lib/utils.py", line 104, in function_to_run
    result = function(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/_universum/modules/vcs/perforce_vcs.py", line 139, in submit_new_change
    raise CriticalCiException("Workspace '" + self.settings.client + "' doesn't exist!")
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

i-keliukh:
Closed by #300.

Issue #271, opened 2019-02-28
No default value in required argument checks


k-dovgan:
## Description

Current argument checks (`if not getattr(self.settings, "user")`) fail if settings were not created via `argparser` and requested parameter is not present in settings at all.

## Steps to reproduce

1. Create settings manually
2. Miss some required option (e.g. "--launcher-config-path")
3. Call `create_component()` and pass created settings to it.

## Expected result

Any kind of "Required parameter not set" message in log (e.g. "Required option CONFIG_PATH (or '--launcher-config-path', or '-lcp') is missing.")

## Actual result

AttributeError: 'settings' object has no attribute <name>

Need to finish the following:
- [x] Commit  partial fixes + tests
- [x] Commit error message fixes
- [x] Commit subcommands fix
- [x] Commit test with del, None and empty string
- [x] Refactor Jenkins server
- [x] Fix argument checks in Jenkins server, add tests
- [x] Fix check_required_option in all code
- [ ] Add all other tests for argument checks

i-keliukh:
The full list of incorrect checking of command line parameters:
```
Found Occurrences  (6 usages found)
    git_vcs.py  (1 usage found)
        247 if not getattr(self.settings, "user") or not getattr(self.settings, "email"):
    jenkins_server.py  (1 usage found)
        29 if getattr(self.settings, name) is None:
    launcher.py  (1 usage found)
        249 if getattr(self.settings, "config_path") is None:
    perforce_vcs.py  (1 usage found)
        330 if not getattr(self.settings, "client"):
    submit.py  (1 usage found)
        39 if getattr(self.settings, "commit_message") is None:
    teamcity_server.py  (1 usage found)
        31 if getattr(self.settings, name) is None:
```

i-keliukh:
```
local_vcs.py  (1 usage found)
    34 if self.settings.source_dir is None:
```

i-keliukh:
Need to finish the following:
- [ ] Commit  partial fixes + tests
- [ ] Commit error message fixes
- [ ] Commit subcommands fix
- [ ] Commit test with del, None and empty string
- [ ] Refactor Jenkins server
- [ ] Fix argument checks in Jenkins server, add tests
- [ ] Fix check_required_option in all code
- [ ] Add all other tests for argument checks

PR #300, opened 2019-05-03
fix(universum): incorrect checks of parameters

i-keliukh:
Jenkins checks are not fixed, because they require
splitting AutomationServer, BaseServer and JenkinsServer
into separate modules for "poll" and "main" modes.

This change is made in scope of task #271.


i-keliukh:
Closes #119.

PR #301, opened 2019-05-06
update(universum): improve error messages for incorrect parameters

i-keliukh:
This change contributes to #271.

### DIFF
--- a/_universum/modules/vcs/perforce_vcs.py
+++ b/_universum/modules/vcs/perforce_vcs.py
@@ -115,9 +115,11 @@ class PerforceSubmitVcs(PerforceVcs, base_vcs.BaseSubmitVcs):
         super(PerforceSubmitVcs, self).__init__(*args, **kwargs)
 
         if not getattr(self.settings, "client", None):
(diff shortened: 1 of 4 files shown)


_universum/modules/vcs/perforce_vcs.py
```diff
@@ -263,7 +263,9 @@ def __init__(self, *args, **kwargs):
         super(PerforceMainVcs, self).__init__(*args, **kwargs)
 
         if not getattr(self.settings, "client", None):
-            raise IncorrectParameterError("P4CLIENT (-p4c) is not specified. It is required for creating workspace.")
+            raise IncorrectParameterError("P4CLIENT (-p4c) is not specified.\n\n"
```
k-dovgan: I think it'd be better to update this message in accordance to other messages in this commit. Like, "perforce workspace (client) is not specified, it is used for bla-bla, please use command line option bla-bla or environment variable bla-bla". Or alternatively start every other error message with env. var. / option required, so that all of them look alike.

PR #305, opened 2019-05-07
refactor(automation server): split module into parts

i-keliukh:
A single automation server was used by two universum modes:
main and poll. This led to incorrect help message
shown and to inability to check respective options
independently.

With this change applied it is now possible to correctly
display help for command-line parameters and check their
correctness.


i-keliukh:
The change introduced test failures.
They have to be fixed.

Issue #174, opened 2019-01-29
Collabsible formatter for jenkins output


k-dovgan:
Originally created on Fri, 22 Jun 2018 13:36:13 +0300

Before heading add some html tag:
`<button class="collapsible">1. Preparing repository</button><div class="content">`
 and after the contents of collapsed section add final `</div>`:
` | └ [Success]  </div>`

See full example:
```
<button class="collapsible">4. Reporting build start</button>
<div class="content">Some information Nowhere to report. Skipping...
└ [Success] </div>
```

k-dovgan:
Originally posted by k.dovgan on Fri, 22 Jun 2018 15:30:58 +0300

Please add all this examples/explanations to plug-in README

k-dovgan:
Originally posted by o.romaniuk on Fri, 22 Jun 2018 15:36:14 +0300

Postpone this task: the reason universum does not send output to jenkins directly, jenkins resives this data through terminal handler.

k-dovgan:
Resolved on Tue, 18 Sep 2018 16:24:28 +0300

PR #324, opened 2019-07-15
feat: add Universum version as log identifier for Jenkins plugin

### LINKED ISSUE #183: Print universum version to build log
Originally created on Tue, 31 Jul 2018 14:41:55 +0300

I propose to print universum version to build log.
It would be useful for reporting bugs.

k-dovgan:
Resolves #183 
Also applies to #174

### DIFF
--- a/universum.py
+++ b/universum.py
@@ -41,6 +41,7 @@ def define_arguments():
 def run(settings):
     result = 0
     main_module = construct_component(settings.main_class, settings)
+    main_module.out.log("Running {} {}".format(__title__, __version__))
 
     finalized = False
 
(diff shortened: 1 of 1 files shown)


universum.py
```diff
@@ -41,6 +41,7 @@ def define_command(klass, command):
 def run(settings):
     result = 0
     main_module = construct_component(settings.main_class, settings)
+    main_module.out.log("Running {} {}".format(__title__, __version__))
```
i-keliukh: Please change to "Starting Universum x.y.z"

miltolstoy:
One more log line like "Finished Universum x.y.z build" in the end is needed.

PR #329, opened 2019-07-23
fix(tests): fail on non-existing environment variable; 'LocalOutput' mention

k-dovgan:


tests/test_vcs.py
```diff
@@ -37,7 +37,10 @@ def test_git_success_command_line_poll_no_changes(stdout_checker, git_server, tm
 
 
 def test_vcs_error_no_vcs_type(capsys):
-    del os.environ["VCS_TYPE"]
+    try:
+        del os.environ["VCS_TYPE"]
```
i-keliukh: Why this change is needed?

k-dovgan: Because if system _does not_ have a default VCS type claimed in environment (unlike our current CI system), this line fails with KeyError. So the variable should be unset _if present._

PR #331, opened 2019-08-14
fix(test): single poll tests failing because httpretty not started yet in init

k-dovgan:


tests/conftest.py
```diff
@@ -72,25 +73,20 @@ def assert_request_was_not_made(query):
             queries.append(request.querystring)
 
     @staticmethod
-    def reset():
-        httpretty.disable()
+    def run_and_collect(call, *args, **kwargs):
```
i-keliukh: The method name should mention assert as well
IMO, "call" should be renamed to "function"

i-keliukh: As an alternative, we can accept just callable and let users of this function handle arguments themselves. Like this:

```python
http_check.run_and_collect(partial(universum.run, git_poll_environment.settings))
```

k-dovgan:
> Did you run pylint?

Yes, but I'm not quite sure I have the same version as CI servers

PR #343, opened 2019-09-25
fix(p4): pass only expected exceptions on file revert

### LINKED ISSUE #215: Error message on cleanning workspace
Originally created on Tue, 20 Nov 2018 15:55:32 +0200

Do not have any error message if we have some error on revert operation
Add message if we have some P4Exception:
```
def clean_workspace(self):
    try:
        self.p4.client = self.client_name
        report = self.p4.run_revert("//...")
        self.p4report(report)
    except P4Exception:
        pass
    self.p4.delete_client(self.client_name)
```

k-dovgan:
Resolves #215

### DIFF
--- a/_universum/modules/vcs/perforce_vcs.py
+++ b/_universum/modules/vcs/perforce_vcs.py
@@ -561,8 +561,10 @@ class PerforceMainVcs(PerforceWithMappings, base_vcs.BaseDownloadVcs):
             self.p4.client = self.client_name
             report = self.p4.run_revert("//...")
             self.p4report(report)
-        except P4Exception:
-            pass
+        except P4Exception as e:
+            if "Client '{}' unknown".format(self.client_name) not in e.value \
+                    and "file(s) not opened on this client" not in e.value:
+                raise
         self.p4.delete_client(self.client_name)
 
     def finalize(self):
--- a/_universum/modules/vcs/perforce_vcs.py
+++ b/_universum/modules/vcs/perforce_vcs.py
@@ -555,17 +555,30 @@ class PerforceMainVcs(PerforceWithMappings, base_vcs.BaseDownloadVcs):
             self.swarm.mappings_dict = self.mappings_dict
 
     @make_block("Cleaning workspace", pass_errors=False)
-    @catch_p4exception(ignore_if="doesn't exist")
     def clean_workspace(self):
-        try:
+
+        def ignore_multiple_p4exceptions(function, ignore_list, *args, **kwargs):
+            try:
+                function(*args, **kwargs)
+            except P4Exception as e:
+                for item in ignore_list:
+                    if item in e.value:
+                        return
+                raise
+
+        def revert():
             self.p4.client = self.client_name
             report = self.p4.run_revert("//...")
             self.p4report(report)
-        except P4Exception as e:
-            if "Client '{}' unknown".format(self.client_name) not in e.value \
-                    and "file(s) not opened on this client" not in e.value:
-                raise
-        self.p4.delete_client(self.client_name)
+
+        with Uninterruptible(self.out.log_exception, is_critical=False) as run:
+            run(ignore_multiple_p4exceptions,
+                revert,
+                ["Client '{}' unknown".format(self.client_name), "file(s) not opened on this client"])
+            run(ignore_multiple_p4exceptions,
+                self.p4.delete_client,
+                ["Client '{}' doesn't exist".format(self.client_name)],
+                self.client_name)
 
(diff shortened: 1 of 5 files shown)


_universum/modules/vcs/perforce_vcs.py
```diff
@@ -561,8 +561,10 @@ def clean_workspace(self):
             self.p4.client = self.client_name
             report = self.p4.run_revert("//...")
             self.p4report(report)
-        except P4Exception:
-            pass
+        except P4Exception as e:
+            if "Client '{}' unknown".format(self.client_name) not in e.value \
+                    and "file(s) not opened on this client" not in e.value:
+                raise
```
i-keliukh: 1. If we are using explicit try...except, we should use it for handling exception from delete_client as well. This would make this function more readable.
2. We should not do raise here. The exception will skip delete_client.

tests/test_p4_revert_unshelved.py
```diff
@@ -131,3 +132,60 @@ def create_file(filename):
...
+    # Prepare SHELVE_CHANGELIST
+    p4.run_edit(perforce_environment.depot)
+    p4_file.write("This is a change.\n")
+    change = p4.fetch_change()
+    change["Description"] = "CL for shelving"
+    shelve_cl = p4.save_change(change)[0].split()[1]
+    p4.run_shelve("-fc", shelve_cl)
+    settings.PerforceMainVcs.shelve_cls = [shelve_cl]
+
+    result = universum.run(perforce_environment.settings)
```
k-dovgan: The problem of using 'universum_runner' instead of all this mess (as in 'test_integration') is that when working in Docker, Universum is launched as root, and therefore Perforce (P4 CLI) in it is executed as root, and default 'revert' works on any files, even read-only, even in read-only directories

tests/utils.py
```diff
@@ -82,6 +82,7 @@ def create_empty_settings(test_type):
 
 class TestEnvironment(object):
     def __init__(self, directory, test_type):
+        self.directory = directory
```
i-keliukh: The name `directory` is not clear.

tests/test_p4_revert_unshelved.py
```diff
@@ -131,3 +133,38 @@ def create_file(filename):
...
+    shelve_cl = p4.save_change(change)[0].split()[1]
+    p4.run_shelve("-fc", shelve_cl)
+
+    settings = perforce_environment.settings
+    settings.Launcher.output = "console"
+    settings.PerforceMainVcs.shelve_cls = [shelve_cl]
+    settings.Launcher.config_path = p4_file.basename
+
+    result = universum.run(settings)
+    perforce_environment.directory.chmod(0777, rec=1)
```
i-keliukh: Please add a comment regarding order of assert and cleanup

tests/test_p4_revert_unshelved.py
```diff
@@ -131,3 +133,38 @@ def create_file(filename):
...
+    perforce_environment.directory.remove(rec=1)
+    assert result != 0
+    stdout_checker.assert_has_calls_with_param("Errors during command execution( \"p4 revert //...\" )")
+    assert "[Errno 13] Permission denied" in capsys.readouterr().err
```
i-keliukh: There is no check of correct handling the p4 revert exception, only the check for the error message.

tests/perforce_utils.py
```diff
@@ -207,6 +207,7 @@ def perforce_workspace(request, perforce_connection, tmpdir):
             for item in remaining_shelves:
                 p4.run_shelve("-dfc", item["change"])
             p4.delete_client("-f", client_name)
+            print "Hello"
```
i-keliukh: What?

tests/test_p4_revert_unshelved.py
```diff
@@ -131,3 +133,44 @@ def create_file(filename):
...
+    settings.PerforceMainVcs.shelve_cls = [shelve_cl]
+    settings.Launcher.config_path = p4_file.basename
+
+    result = universum.run(settings)
+    # Clean up the directory at once to make sure it doesn't remain non-writable even if some assert fails
+    perforce_environment.temp_dir.chmod(0777, rec=1)
+    perforce_environment.temp_dir.remove(rec=1)
+
+    assert result != 0
+    # Check that both 'revert' and 'clean workspace' operations were executed unsuccessfully
```
i-keliukh: The comment emphasizes unsuccessful execution, while the important part is that **both** operations were executed.
In fact, you'd better mention all three operations in the first comment, because execution of all of them is important.

PR #347, opened 2019-10-01
feat(jenkins_plugin): increase version, update gitignore with plugin-specific files

miltolstoy:
feat(jenkins_plugin): increase version, update gitignore with plugin-specific files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,8 @@ htmlcov
 temp
 artifacts
 temp_config.py
+.classpath
+.factorypath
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -11,3 +11,8 @@ htmlcov
 temp
 artifacts
 temp_config.py
+.classpath
```
i-keliukh: Doesn't it make more sense to add this to plugin-specific .gitignore?

PR #350, opened 2019-10-03
fix(universum): handle SIGTERM properly

### LINKED ISSUE #272: Wrong processing of execution interruption
## Description

Current implementation of enforced program stop interception doesn't work, because all measures to prevent this are defined after execution the code that is interrupted.
To avoid this, `signal.signal` and `atexit` calls must be done before try-except block with `Uninterruptible`.

## Steps to reproduce

1. Launch Universum in any mode
2. Send SIGTERM to Universum process, or press ctrl+C before execution is finished

## Expected result

`finalize()` is executed anyway.

## Actual result

Program exits right away.

k-dovgan:
Resolves #272

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -2,6 +2,11 @@
 # -*- coding: UTF-8 -*-
 
 import os
+import signal
+import time
+
+import pytest
+import sh
 
 
 def get_line_with_text(text, log):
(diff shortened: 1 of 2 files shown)


tests/test_integration.py
```diff
@@ -319,3 +324,32 @@ def test_environment(universum_runner):
 configs = upper * lower
 """)
     assert "This string should be in log" in log
+
+
+@pytest.mark.parametrize("name,terminate_type",
```
i-keliukh: Use ids=[...] instead of fake parameter "name"

PR #351, opened 2019-10-04
fix(jenkins_plugin): fix log displaying at job terminating

### LINKED ISSUE #348: Incorrect formatting of logs on job interruption
## Description

Jenkins log collapser plugin produces weird results if execution is interrupted.
See example here: https://jenkins.itnews.biz.ua/job/Universum/job/Check%20commit/84/console

miltolstoy:
# Description
fix(jenkins_plugin): fix log displaying at job terminating

There is no specific log line, which can indicate that job is stopped unexpectedly and following log is not Universum log. That's why skipping log processing, if some log without indentation appears in place were it should be with indentation.

Normal log example:
```
  |  step data...
  |  step data...
  |  step data...
```

Terminated log example:
```
  |  step data...
  |  step data...
Terminated
Build was aborted
```

Resolves #348 

# How Has This Been Tested?
Interrupt Jenkins build with Universum inside in various configuration and time moments.

### DIFF
--- a/universum_log_collapser/pom.xml
+++ b/universum_log_collapser/pom.xml
@@ -9,7 +9,7 @@
     </parent>
     <groupId>io.jenkins.plugins</groupId>
     <artifactId>universum_log_collapser</artifactId>
-    <version>1.1</version>
+    <version>1.2</version>
     <packaging>hpi</packaging>
     <properties>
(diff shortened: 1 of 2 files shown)


k-dovgan:
Related to #350

Issue #358, opened 2019-10-11
Add Jenkins plugin Java sources testing


miltolstoy:
Add Jenkins plugin Java sources testing

PR #360, opened 2019-10-17
test(jenkins_plugin): add Java tests for log start, end, break

miltolstoy:
# Description
1. Java tests for log start, end and break added.
2. Make regexps for log end more specific.
3. Move processing of Jenkins log end to top of the function, because it was skipped as "non-Universum log".

Relates to #358


k-dovgan:
Please add executing Java tests to `configs.py` to see test result immediately

miltolstoy:
> The problem is, in current implementation Maven tests pull all required packages to working directory. This directory is obviously cleaned up for CI purposes. And pulling takes a lot of time, yet along traffic.
> I cannot approve adding these tests without proper dependency management

Maven local repository is located in ~/.m2. It can be changed by editing /etc/maven/settings.xml.
https://www.baeldung.com/maven-local-repository

PR #367, opened 2020-01-06
test(jenkins_plugin): create JS tests on static HTML

miltolstoy:


universum_log_collapser/e2e/static_html.test.js
```diff
@@ -0,0 +1,90 @@
...
+});
+
+
+test("pipeline issue fixed", async() => {
+    let pipelineElementText = await page.$eval(".pipeline-node-7", e => e.innerHTML);
+    expect(pipelineElementText).toHaveLength(0);
+});
+
+
+async function waitForHandledResults(count) {
```
i-keliukh: IMO, it would be logical to move this function to the beginning of the file, closer to the usage

miltolstoy: I prefer to locate all helper functions in the end of the file. I hope that this allow easier reading of more important functions (tests preconditions and tests in this case) without mixing them with "low-level" functions.

i-keliukh:
This PR is removed from master using force push

PR #383, opened 2020-01-30
test(jenkins_plugin): add JS tests on generated HTML

miltolstoy:
Resolves #(359)

### DIFF
--- a/universum_log_collapser/e2e/live_log_html.test.js
+++ b/universum_log_collapser/e2e/live_log_html.test.js
@@ -1,0 +1,27 @@
+const utils = require("./utils");
+
+beforeAll(async () => {
+    await utils.openPage("generated.html");
+    await utils.waitForHandledResults(2);
+});
+
+test('failed steps class changed', async () => {
+    await utils.checkFailedStepsHandled();
+});
+
+test("failed steps labels colored", async() => {
+    let labelColors = await utils.getLabelsColors();
+    expect(labelColors).toHaveLength(15);
+    await utils.checkRedLabels(labelColors, [5, 7, 8, 10])
+});
+
+test("steps labels have no spans", async() => {
+    let spanColors = await utils.getLabelSpansColors();
+    expect(spanColors).toHaveLength(0);
+});
+
+test("failed steps expanded", async() => {
+    let labelsChecked = await utils.getLabelsCheckedState();
+    expect(labelsChecked).toHaveLength(15);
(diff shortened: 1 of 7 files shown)


universum_log_collapser/e2e/static_html.test.js
```diff
@@ -1,92 +1,34 @@
+const utils = require("./utils");
+
 beforeAll(async () => {
-    const path = require('path');
-    const currDir = path.dirname(module.filename);
-    await page.goto('file://' + currDir + '/test.html');
-    await waitForHandledResults(2);
+    await utils.openPage("test.html");
+    await utils.waitForHandledResults(2);
```
k-dovgan: It would be nice to declare a constant somewhere, describing that '2' is a current count of failed steps in sample output, and use it instead of simply passing magical '2' in every function usage. Using bare number is not only confusing, but is less convenient in case of changes to sample output

universum_log_collapser/e2e/static_html.test.js
```diff
@@ -1,92 +1,35 @@
+const utils = require("./utils");
+
 beforeAll(async () => {
-    const path = require('path');
-    const currDir = path.dirname(module.filename);
-    await page.goto('file://' + currDir + '/test.html');
-    await waitForHandledResults(2);
+    await utils.openPage("test.html");
+    const expFailedStepsCount = 2; // depends on test HTML content
```
k-dovgan: Are you sure you want to declare the exact same constant several times instead of just sharing it?

universum_log_collapser/e2e/live_log_html.test.js
```diff
@@ -0,0 +1,28 @@
...
+    await utils.openPage("generated.html");
+    const expFailedStepsCount = 2; // depends on Universum test configuration
+    await utils.waitForHandledResults(expFailedStepsCount);
+});
+
+test('failed steps class changed', async () => {
+    await utils.checkFailedStepsHandled();
+});
+
+test("failed steps labels colored", async() => {
```
i-keliukh: Not a big deal, but the formatting of () is inconsistent: sometimes you add a space before and sometimes you don't.

universum_log_collapser/e2e/static_html.test.js
miltolstoy: Yes.
It's just a coincidence that 2 different HTMLs have same failed steps count. They are not related to each other in any way.
I can update one of HTMLs to have 3 failed steps to avoid embarrassment.

PR #390, opened 2020-04-03
fix(p4): delete CLs too in case of force clean

### LINKED ISSUE #389: --p4-force-clean doesn't delete CLs
## Description
If cleaned workspace contains CLs, it is not deleted despite `--p4-force-clean` option passed

## Steps to reproduce
1. Create a workspace
2. Create a CL in this workspace
3. Launch Universum with `--vcs-type p4 --p4-force-clean --p4-client <created workspace name>` parameters

## Expected result
Workspace deleted and recreated empty.

## Actual result
Workspace not deleted; exception logged and Universum does not continue execution.

k-dovgan:
Resolves #389

### DIFF
--- a/_universum/modules/vcs/perforce_vcs.py
+++ b/_universum/modules/vcs/perforce_vcs.py
@@ -560,15 +560,15 @@ class PerforceMainVcs(PerforceWithMappings, base_vcs.BaseDownloadVcs):
             self.p4.client = self.client_name
             report = self.p4.run_revert("//...")
             self.p4report(report)
-            shelves = p4.run_changes("-c", self.client_name, "-s", "shelved")
+            shelves = self.p4.run_changes("-c", self.client_name, "-s", "shelved")
             for item in shelves:
                 self.out.log("Deleting shelve from CL " + item["change"])
-                p4.run_shelve("-d", "-c", item["change"])
+                self.p4.run_shelve("-d", "-c", item["change"])
             self.p4.run_revert("//...")
-            all_cls = p4.run_changes("-c", self.client_name, "-s", "pending")
+            all_cls = self.p4.run_changes("-c", self.client_name, "-s", "pending")
             for item in all_cls:
                 self.out.log("Deleting CL " + item["change"])
-                p4.run_change("-d", item["change"])
+                self.p4.run_change("-d", item["change"])
         except P4Exception as e:
             if "Client '{}' unknown".format(self.client_name) not in e.value \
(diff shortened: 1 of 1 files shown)


_universum/modules/vcs/perforce_vcs.py
```diff
@@ -560,6 +560,15 @@ def clean_workspace(self):
             self.p4.client = self.client_name
             report = self.p4.run_revert("//...")
             self.p4report(report)
+            shelves = p4.run_changes("-c", self.client_name, "-s", "shelved")
+            for item in shelves:
+                self.out.log("Deleting shelve from CL " + item["change"])
+                p4.run_shelve("-d" , "-c", item["change"])
+            self.p4.run_revert("//...")
```
testing-dev, line 567: I this revert really needed? As far as I understand you have already reverted repo in the line 561.

k-dovgan: Yes. 'p4 shelve -d -c 1234' will delete any changes, stored in CL 1234 _after_ unshelving these changes into CL itself. That's why _before_ this command CL contents must be already reverted to prevent resolving; and that's why this CL should be reverted again after.
This can be avoided by using '-f' option, but it is not available to non-admin users.

testing-dev: Are you sure? 
I can remove the shelve without any "auto-unshelving" in the GUI version of P4. Command is the same "p4 shelve -d -c 18381526"

Application: 
Helix Visual Client version: P4V/NTX64/2019.2/1904275

k-dovgan: For more information please refer to official guide:
https://www.perforce.com/manuals/v18.1/cmdref/index.html#CmdRef/p4_shelve.html
('-d' option)

testing-dev: Ok, thank you

i-keliukh: I just checked this behavior on real perforce instance.
There is no issue with having the same file in pending state (being checked out) while deleting its shelved version.
`p4 shelve -d -c 1234` correctly deleted shelve and didn't change the checked out file.

i-keliukh: The decision to keep second revert has to be documented somewhere. This is because the p4 documentation is counterintuitive and doesn't correspond to implementation.

PR #397, opened 2020-04-10
fix: tests

i-savynikh:
reason: python3 uses UTF-8 by default.
'#!/usr/bin/env python3' not requires to run Universum but cause issue
when file saved with 'windows' line ending.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-# -*- coding: UTF-8 -*-
 
 from __future__ import absolute_import
 from __future__ import print_function
@@ -45,7 +43,8 @@ setup(
         'glob2',
         'requests',
         'sh',
-        'lxml'
+        'lxml',
+        'six',
     ],
     extras_require={
         'docs': [docs],
--- a/_universum/modules/launcher.py
+++ b/_universum/modules/launcher.py
@@ -4,9 +4,9 @@ from __future__ import absolute_import
 import os
 import re
 import sys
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -20,8 +18,8 @@ def readme():
 )
 
 vcs = (
-    'gitpython>=2.1.8',
-    'P4'
+    'gitpython>=3.0.5',
+    'p4python>=2019.1'
```
i-keliukh, line 22: Dows P4 package have different names between python 2 and 3?

tests/configs.py
```diff
@@ -14,16 +14,18 @@ def run_virtual(cmd):
 
 configs = Variations([dict(name="Update Docker images", command=["make", "images"]),
                       dict(name="Create virtual environment",
-                           command=["python3.7", "-m", "venv", env_name, "--system-site-packages"]),
```
i-keliukh, line 17: What is wrong with reusing system packages?

setup.py
i-savynikh: yes, the correct way to install p4python now is a **pip install p4python**. It's work OK for python3.7 now.

tests/configs.py
i-savynikh: system have some pre-installed packages. Universum can have dependencies which aren't defined in setup.py but installed on CI system. In this case we will not see error from CI but it really is.
virtualenv will use 'system' cache to install packages in case cache still actual.

Issue #396, opened 2020-04-10
Add submitter tests for unsuccessful submit


k-dovgan:
List of scenarios that should be tested and handled properly:
- [ ] requirements of special CL description format not fulfilled
- [ ] requirements of mandatory code review not fulfilled
- [ ] no 'write' rights for the target
- [ ] submitted files have been changed since CL creation (e.g. by another user)
- [ ] server-side problems (e.g. server application crashed)
- [ ] network problems (e.g. server temporarily inaccessible)

k-dovgan:
3&4 implemented in #398

PR #398, opened 2020-04-14
add(test): submit to protected branches

### LINKED ISSUE #388: Created CL is not cleaned on failed submit
## Description
If submit fails, reconciled changes are stored in newly created CL; this behaviour interferes with automation.

### Steps to reproduce
1. Configure special p4 submit requirements, such as mandatory review or explicit CL description.
2. Use Universum to submit a CL not meeting those requirements.

### Expected result
Exception logged; local files left unchanged; created CL deleted; changes in VCS reverted

### Actual result
Exception logged; local files left unchanged; created CL left hanging with all changes added for VCS

k-dovgan:
Check 3&4 from #396

### DIFF
--- a/tests/perforce_utils.py
+++ b/tests/perforce_utils.py
@@ -205,7 +205,7 @@ def perforce_workspace(request, perforce_connection, tmpdir):
         p4.save_protect(permissions)
 
         triggers = {'Triggers': [
-            'test.check change-submit //depot/trigger-protected/... "/bin/cp . ."'
+            'test.check change-submit //depot/trigger-protected/... "false"' # trigger to prevent any submits to this branch
         ]}
         p4.save_triggers(triggers)
 
--- a/tests/perforce_utils.py
+++ b/tests/perforce_utils.py
@@ -198,9 +198,8 @@ def perforce_workspace(request, perforce_connection, tmpdir):
         permissions['Protections'] = [
             'write user * * //...',
             'list user * * -//spec/...',
-            'super user p4user * //...',                          # first three rows are default
-            '=write user p4user * -//depot/write-protected/...',  # prohibit p4user to submit changes to this branch
-            '=open user p4user * -//depot/open-protected/...'     # prohibit p4user to check out any files in this branch
+            'super user p4user * //...',                         # first three rows are default
+            '=write user p4user * -//depot/write-protected/...'  # prohibit p4user to submit changes to this branch
         ]
         p4.save_protect(permissions)
 
--- a/tests/perforce_utils.py
+++ b/tests/perforce_utils.py
@@ -196,8 +196,10 @@ def perforce_workspace(request, perforce_connection, tmpdir):
 
         permissions = p4.fetch_protect()
         permissions['Protections'] = [
+            'write user * * //...',
+            'list user * * -//spec/...',
             'super user p4user * //...',
-            'open user p4user * -//spec/...'
+            '=write user p4user * -//depot/protected/...'
         ]
         p4.save_protect(permissions)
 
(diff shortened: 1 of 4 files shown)


tests/perforce_utils.py
```diff
@@ -194,6 +194,21 @@ def perforce_workspace(request, perforce_connection, tmpdir):
         change["Description"] = "Test submit"
         p4.run_submit(change)
 
+        permissions = p4.fetch_protect()
+        permissions['Protections'] = [
+            'write user * * //...',
```
i-keliukh, line 199: Are the first three items default?
If yes, please explicitly mark them in comment.

tests/perforce_utils.py
```diff
@@ -194,6 +194,21 @@ def perforce_workspace(request, perforce_connection, tmpdir):
...
+            'list user * * -//spec/...',
+            'super user p4user * //...',
+            '=write user p4user * -//depot/write-protected/...',
```
i-keliukh, line 202: These two lines need explanation.

tests/perforce_utils.py
```diff
@@ -194,6 +194,21 @@ def perforce_workspace(request, perforce_connection, tmpdir):
...
+            '=open user p4user * -//depot/open-protected/...'
+        ]
+        p4.save_protect(permissions)
+
+        triggers = {'Triggers': [
+            'test.check change-submit //depot/review-protected/... "/bin/cp . ."'
```
i-keliukh, line 208: The intention of using "/bin/cp . ." is not clear

tests/test_p4_submit.py
```diff
@@ -37,3 +38,23 @@ def test_success_changing_checked_out_file(p4_submit_environment):
     p4_submit_environment.p4.run_submit(change)
 
     assert p4_submit_environment.file_present(str(target_file))
+
+
+@pytest.mark.parametrize("test_type,expected",
+                         [("open", 0), ("write", 1), ("review", 1)],
```
i-keliukh, line 44: "review" doesn't describe the protection exactly.
I would say it is trigger-protected branch.

tests/test_p4_submit.py
```diff
@@ -37,3 +38,23 @@ def test_success_changing_checked_out_file(p4_submit_environment):
     p4_submit_environment.p4.run_submit(change)
 
     assert p4_submit_environment.file_present(str(target_file))
+
+
+@pytest.mark.parametrize("test_type,expected",
```
i-keliukh, line 43: Please change the `test_type` to `branch` and its values to "open-protected", "write-protected", etc.

tests/test_p4_submit.py
i-keliukh, line 44: It is not clear why `open` succeeds, but write and review fail. This has to be documented somewhere.

tests/test_p4_submit.py
```diff
@@ -37,3 +38,23 @@ def test_success_changing_checked_out_file(p4_submit_environment):
...
+    file_to_add.write(text + "\n")
+
+    settings = copy.deepcopy(p4_submit_environment.settings)
+    setattr(settings.Submit, "reconcile_list", [unicode(file_to_add)])
+
+    result = universum.run(settings)
+    assert result == expected
+
+    p4 = p4_submit_environment.p4
+    assert not p4.run_changes("-c", p4_submit_environment.client_name, "-s", "pending")
```
i-keliukh, line 59: Please document what these asserts check exactly.
Otherwise, the reader has to google for p4 documentation.

tests/perforce_utils.py
k-dovgan: This will try to launch any executable on server side. This is just an executable granted to exist and parameters granted to fail. The alternative is to commit an always-failing script to repo, but I didn't think it was actually needed.

i-keliukh: The more clear way is to use `false` command

PR #402, opened 2020-04-17
fix(ci): config didn't verify dependencies correctly

i-savynikh:
'--system-site-packages' usage can lead to misscheck of dependencies defined in setup.py. For more information see appropriate flag documentation.
'upgrade' flag requires to test Universum using latest versions of packages.


k-dovgan:
You note 'upgrade' in comment, but use '-U' in code. It's a little bit misleading. Also the explanation does not explain what '-U' does exactly and why do we need it. And I see nothing about system cache in comment.

PR #406, opened 2020-04-21
fix(p4): reverted files may no longer be accessible

k-dovgan:
This change fixes the rare bug, happening when files to be reverted by `finalize()` function are no longer accessible by P4. This might be a result of two scenarios:
1. actions done by Universum steps (e.g. `chmod` command in scripts)
2. desirable workspace name passed to Universum corresponds to a workspace that already exists and user running the Universum has no right to change files to be reverted in this workspace

### DIFF
--- a/tests/test_p4_exception_handling.py
+++ b/tests/test_p4_exception_handling.py
@@ -1,0 +1,67 @@
+# -*- coding: UTF-8 -*-
+# pylint: disable = redefined-outer-name
+
+import pytest
+
+import universum
+from tests.perforce_utils import P4Environment
+
+
+@pytest.fixture()
+def perforce_environment(perforce_workspace, tmpdir):
+    yield P4Environment(perforce_workspace, tmpdir, test_type="main")
+
+
+def test_p4_forbidden_local_revert(perforce_environment, capsys):
+    p4 = perforce_environment.p4
+    p4_file = perforce_environment.repo_file
+
+    config = """
+from _universum.configuration_support import Variations
+
+configs = Variations([dict(name="Restrict changes", command=["chmod", "-R", "555", "."]),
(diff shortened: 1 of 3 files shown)


tests/test_p4_exception_handling.py
```diff
@@ -0,0 +1,66 @@
...
+
+    result = universum.run(settings)
+    # Clean up the directory at once to make sure it doesn't remain non-writable even if some assert fails
+    perforce_environment.temp_dir.chmod(0777, rec=1)
+    perforce_environment.temp_dir.remove(rec=1)
+
+    assert result != 0
+
+    assert "[Errno 13] Permission denied" in capsys.readouterr().err
+    # make sure submitter didn't leave any pending CLs in the workspace
```
i-keliukh, line 44: Submitter?

tests/test_p4_exception_handling.py
```diff
@@ -0,0 +1,66 @@
...
+    assert not p4.run_changes("-c", perforce_environment.client_name, "-s", "pending")
+    # make sure submitter didn't leave any pending changes in default CL
+    assert not p4.run_opened("-C", perforce_environment.client_name)
+
+
+def test_p4_print_exception(perforce_environment, stdout_checker):
+    p4 = perforce_environment.p4
+    client = p4.fetch_client(perforce_environment.client_name)
+    client["Options"] = "noallwrite noclobber nocompress locked nomodtime normdir"
+    p4.save_client(client)
```
i-keliukh, line 54: Are you sure this test checks the finalize?
It looks like it detects failures of the initial workspace cleanup.

k-dovgan: Does it matter? We need to check if exceptions are shown at all, no matter where

PR #414, opened 2020-04-29
fix(test): nonci test started to fail due to 'sh'

i-keliukh:
The reason of the failure is that 'sh' changed the format of the
command line string returned for the launched process, and the
test is designed to check exactly one format.

Fixed the failure by avoiding checking of the command line
presence in logs at all.

Also, improved the test to check more features of the nonci mode.

### DIFF
--- a/tests/test_nonci.py
+++ b/tests/test_nonci.py
@@ -5,39 +5,62 @@
 config = """
 from _universum.configuration_support import Variations
 
-configs = Variations([dict(name="test_step", artifacts="test_nonci.txt",
-                           command=["bash", "-c", '''echo "test nonci" > test_nonci.txt'''])])
+configs = Variations([dict(name="artifact check",
+                           command=["bash", "-c", '''cat /artifacts/test_nonci.txt''']),
+#                                                    ^ this helps to check artifact is deleted before launch 
+
+                      dict(name="test_step", artifacts="test_nonci.txt",
+                           command=["bash", "-c", '''echo "pwd:[$(pwd)]" && echo "test nonci" > test_nonci.txt'''])])
+#                                                    ^ this helps to check the path is not changed
 """
 
 
 def test_launcher_output(universum_runner_nonci):
+    """
+    This test verifies that nonci mode changes the behavior of the universum by checking its output to console and log
+    files. Specifically, it checks the following features of the nonci mode:
+     - default output of the step is console, even when running in the terminal
+     - sources are not copied to temp directory
+     - artifacts are deleted before launching configs
+     - version control and review system are not used
+    """
     file_output_expected = "Adding file /artifacts/test_step_log.txt to artifacts"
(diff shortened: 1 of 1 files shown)


tests/test_nonci.py
```diff
@@ -5,39 +5,62 @@
...
 
-    console_out_log = universum_runner_nonci.run(config) # defult -lo is console
-    assert file_output_expected not in console_out_log
-    assert step_log_expected in console_out_log
-    # nonci doesn't required to clean artifacts between calls
+    console_out_log = universum_runner_nonci.run(config)
+    assert "Cleaning artifacts..." in console_out_log           # nonci cleans artifacts on project launch
+    assert "Old artifact" not in console_out_log                # the artifacts are actually deleted
+    assert file_output_expected not in console_out_log          # nonci doesn't write logs to the file by default
+    assert "Reporting build start" not in config                # nonci doesn't report build start
```
k-dovgan, line 36: I think it should be `console_out_log`, not `config`

tests/test_nonci.py
```diff
@@ -5,39 +5,63 @@
...
+    universum_runner_nonci.environment.assert_successful_execution(
+        "bash -c 'mkdir /artifacts; echo \"Old artifact\" > /artifacts/test_nonci.txt'")
 
-    console_out_log = universum_runner_nonci.run(config) # defult -lo is console
-    assert file_output_expected not in console_out_log
-    assert step_log_expected in console_out_log
-    # nonci doesn't required to clean artifacts between calls
+    console_out_log = universum_runner_nonci.run(config)
+    assert "Cleaning artifacts..." in console_out_log           # nonci cleans artifacts on project launch
+    assert "Old artifact" not in console_out_log                # the artifacts are actually deleted
```
i-savynikh, line 35: I can't found "Old artifact" string in a project. Does this check able to fail?

i-keliukh: It is located on line 31

tests/test_nonci.py
```diff
@@ -5,39 +5,63 @@
...
+
+    universum_runner_nonci.environment.assert_successful_execution(
+        "bash -c 'mkdir /artifacts; echo \"Old artifact\" > /artifacts/test_nonci.txt'")
 
-    console_out_log = universum_runner_nonci.run(config) # defult -lo is console
-    assert file_output_expected not in console_out_log
-    assert step_log_expected in console_out_log
-    # nonci doesn't required to clean artifacts between calls
+    console_out_log = universum_runner_nonci.run(config)
+    assert "Cleaning artifacts..." in console_out_log           # nonci cleans artifacts on project launch
```
i-savynikh, line 34: Please divide this block on two: check "lines expected in logs" and "shall not be in logs"
In my opinion it will be more readable.
```
assert line in log
assert line in log

assert line not in log
...
```

i-keliukh: done

PR #420, opened 2020-05-04
fix(git): git module changed API in python 3

i-keliukh:
According to the documentation, to change config, user must
either explicitly call 'release' method or use context
managers.

### DIFF
--- a/_universum/modules/vcs/git_vcs.py
+++ b/_universum/modules/vcs/git_vcs.py
@@ -267,9 +267,9 @@ class GitSubmitVcs(GitVcs):
         except git.exc.InvalidGitRepositoryError:
             raise CriticalCiException("'" + self.settings.project_root + "' does not contain a Git repository")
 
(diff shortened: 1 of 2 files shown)


_universum/modules/vcs/git_vcs.py
```diff
@@ -267,9 +267,9 @@ def git_commit_locally(self, description, file_list, edit_only=False):
         except git.exc.InvalidGitRepositoryError:
             raise CriticalCiException("'" + self.settings.project_root + "' does not contain a Git repository")
 
-        configurator = self.repo.config_writer()
-        configurator.set_value("user", "name", self.settings.user)
-        configurator.set_value("user", "email", self.settings.email)
+        with self.repo.config_writer() as configurator:
+            configurator.set_value("user", "name", self.settings.user)
+            configurator.set_value("user", "email", self.settings.email)
```
i-savynikh, line 272: looks like this chunk need a separate function

PR #424, opened 2020-05-06
fix(test): code report test only checked pylint analyzer

i-keliukh:
Code report test contained checks only for verifying pylint analyzer
module, but it didn’t check the functionality of code report at all.
After fixing the test, it became clear that pylint analyzer doesn’t
produce any reports if there are no isseus. However, code report
always expects reports to be present. Fixed the pylint analyzer to
always produce report, even if there are no issues found.

Also removed code report tests that are producing output to stdout
instead of writing it to files, because code report module cannot
handle such output.

### DIFF
--- a/analyzers/pylint.py
+++ b/analyzers/pylint.py
@@ -70,8 +70,7 @@ class PylintAnalyzer:
             utils.analyzers_output(self.settings.result_file, issues_loads)
             if issues_loads:
                 return 1
-            else:
-                return 0
+            return 0
 
         except ValueError as e:
             sys.stderr.write(str(e))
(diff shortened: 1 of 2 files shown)


analyzers/pylint.py
```diff
@@ -72,10 +67,11 @@ def _handle_pylint_output(self, issues: str) -> int:
                 issue["message"] = issue["message"].replace("&lt;", "<").replace("&gt;", ">").replace("&amp;", "&")
                 issues_loads.append(issue)
 
+            utils.analyzers_output(self.settings.result_file, issues_loads)
             if issues_loads:
```
i-savynikh, line 71: In my opinion following code looks better: `return 1 if issue_loads else 0`

k-dovgan: In my opinion multi-line if looks more clear, I'd just remove extensive `else`

i-keliukh:
Force pushed as commit f8cbfe45a5c5aed37009669fab26c41c84a4252f to fix typo in the commit message

PR #429, opened 2020-05-07
fix(test): only init vcs in tests that need it

i-keliukh:


tests/deployment_utils.py
```diff
@@ -301,5 +301,7 @@ def nonci(request):
 
 
 def pytest_generate_tests(metafunc):
-    if metafunc.definition.get_closest_marker('nonci_applicable'):
-        metafunc.parametrize('nonci', (False, True), ids=('no subcmd', 'subcmd: nonci',))
+    if metafunc.definition.get_closest_marker('nonci_and_main'):
```
k-dovgan, line 304: Please add a comment, explaining that 'main_only' is a default and doesn't need to be marked

PR #444, opened 2020-05-26
fix(jenkins): remove unnecessary return

k-dovgan:


universum/modules/automation_server/jenkins_server.py
```diff
@@ -38,13 +35,9 @@ def trigger_build(self, revision):
...
-            six.moves.urllib.request.urlopen(processed_url)
-        except six.moves.urllib.error.URLError as url_error:
-            raise CriticalCiException(f"Error opening URL, error message {url_error.reason}")
-        except ValueError as value_error:
-            raise CriticalCiException(f"Error opening URL, error message {value_error}")
-
-        return True
+            requests.get(processed_url)
+        except (requests.HTTPError, requests.ConnectionError, ValueError) as e:
+            raise CriticalCiException(f"Error opening URL, error message {e}")
```
i-keliukh, line 40: Previously it user reason field. 
Why it is not needed anymore?

k-dovgan: Because in Python3 there's no reason field, the text is in repr itself.

PR #448, opened 2020-05-29
feat(github-handler): add a new Universum mode

k-dovgan:


universum/modules/vcs/github_vcs.py
```diff
@@ -17,7 +21,34 @@ def get_time():
...
 
-class GithubMainVcs(ReportObserver, git_vcs.GitMainVcs):
+class GithubToken(Module):
+
+    @staticmethod
+    def define_arguments(argument_parser):
+        parser = argument_parser.get_or_create_group("GitHub", "GitHub repository settings")
+
+        parser.add_argument("--github-integration", "-ght", dest="integration_id", metavar="GITHUB_INTEGRATION",
+                            help="GitHub application ID (see 'general')")
```
i-keliukh, line 31: What does it refer to?
Where is that 'general'?

universum/modules/vcs/github_vcs.py
```diff
@@ -17,7 +21,34 @@ def get_time():
...
+        parser.add_argument("--github-installation", "-ghs", dest="installation_id", metavar="GITHUB_INSTALLATION",
+                            help="in-project installation ID (see 'integrations & services')"
```
i-keliukh, line 33: Are you sure "in-project" is correct term?
I think github uses "repository" in the same context.

universum/modules/vcs/github_vcs.py
```diff
@@ -17,7 +21,34 @@ def get_time():
...
+                                 " - not needed for 'github-handler'")
```
i-keliukh, line 34: If the handler doesn't require this parameter we need to remove it.
Note we already have number of modes of operation with excessive parameters.
We'd better not increase this number further.

k-dovgan: ???
GitHub Handler is not the only user of this arguments
While Handler can get the installation ID out of payload, Universum should still receive it for report for review somehow

universum/modules/vcs/github_vcs.py
k-dovgan: GitHub documentation for Apps. I'd rather fix documentation in next commit, these lines are just copied from previous implementation. I can remove help completely from this commit though for less confusion. (Also applies for the above)

universum/modules/vcs/github_vcs.py
```diff
@@ -79,7 +96,7 @@ def __init__(self, *args, **kwargs):
         self.clone_url = six.moves.urllib.parse.urlunsplit(parsed_repo)
         self.headers = {
             "Accept": "application/vnd.github.antiope-preview+json",
-            "Authorization": "token " + self.settings.token
+            "Authorization": "token " + self.get_token()
```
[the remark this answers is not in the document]
k-dovgan: Sure, it is just a different part of the refactoring and should be reimplemented in another way - in another commit. For now it simply does the very same thing it did before

universum/modules/vcs/github_vcs.py
```diff
@@ -17,7 +21,34 @@ def get_time():
...
+        parser.add_argument("--github-private-key", "-ghk", dest="key_path", metavar="GITHUB_PRIVATE_KEY",
+                            help="Application private key file path")
+
+    def __init__(self, *args, **kwargs):
+        super(GithubToken, self).__init__(*args, **kwargs)
+        with open(self.settings.key_path) as f:
+            private_key = f.read()
+        self.integration = github.GithubIntegration(self.settings.integration_id, private_key)
+
+    def get_token(self, installation_id=None):
```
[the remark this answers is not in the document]
k-dovgan: See comment for the next line. It should and will be changed, just not in this commit.

universum/modules/vcs/github_vcs.py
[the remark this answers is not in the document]
i-keliukh: Ok, then I'm resolving both comments.

universum/modules/vcs/github_vcs.py
i-keliukh: I looked there before writing my comment. Could you, please, provide direct link?
Also, considering these are completely new parameters, where you could copy them from?

universum/modules/vcs/github_vcs.py
i-keliukh: Then we need to split GithubToken class

setup.py
```diff
@@ -35,7 +35,8 @@ def readme():
         'requests',
         'sh',
         'lxml',
-        'six'
+        'six',
+        'pygithub'
```
i-keliukh, line 39: We do not make VCS modules hard dependencies, but we've made github a hard dependency.
This has to be consistent with current approach.

universum/modules/automation_server/jenkins_server.py
```diff
@@ -18,9 +20,12 @@ class JenkinsServerForTrigger(BaseServerForTrigger):
     @staticmethod
     def define_arguments(argument_parser):
         parser = argument_parser.get_or_create_group("Jenkins variables", "Jenkins-specific parameters")
-        parser.add_argument('--jenkins-trigger-url', '-jtu', dest='trigger_url',
-                            help='Url to trigger, must include exactly one conversion specifier (%%s) to be '
-                                 'replaced by CL number, for example: http://localhost/%%s', metavar="URL")
+        parser.add_argument('--jenkins-trigger-url', '-jtu', dest='trigger_url', metavar="TRIGGER_URL",
+                            help='Url to trigger, ending with "build" or "buildWithParameters" accordingly')
+        parser.add_argument('--jenkins-trigger-parameter', "-jtp", action="append", nargs='+', dest="param_list",
```
i-keliukh, line 25: Why we need to pass parameter explicitly?
User can already include them into trigger url.

universum/modules/automation_server/jenkins_server.py
```diff
@@ -29,13 +34,20 @@ def __init__(self, *args, **kwargs):
...
                                           "Please specify the url by using '--jenkins-trigger-url' ('-jtu')\n"
                                           "command-line option or URL environment variable.")
+        self.params = unify_argument_list(self.settings.param_list)
 
-    def trigger_build(self, revision):
-        processed_url = self.settings.trigger_url % revision
+    def trigger_build(self, param_dict=None):
+        processed_url = self.settings.trigger_url
+        if param_dict:
+            self.params.append(urllib.parse.urlencode(param_dict))
```
i-keliukh, line 42: `self.params` will grow larger on each call to `trigger_build`, which is counter intuitive.
Constructed object must keep it initial state.

universum/github_handler.py
```diff
@@ -0,0 +1,61 @@
...
+    @staticmethod
+    def define_arguments(argument_parser):
+        argument_parser.add_argument('--event', '-e', dest='event', metavar="GITHUB_EVENT",
+                                     help='Currently parsed from "x-github-event" header')
+        argument_parser.add_argument('--payload', '-p', dest='payload', metavar="PAYLOAD",
+                                     help='<actual help coming later> leave "-" to read from stdin')
+
+    def __init__(self, *args, **kwargs):
+        super(GithubHandler, self).__init__(*args, **kwargs)
+        if self.settings.payload == '-':
```
i-keliukh, line 28: This is just a reminder: we agreed to use curl-like approach:
- `-` to read from stdin
- `@` to read from file
- no special char to pass data directly

universum/github_handler.py
```diff
@@ -0,0 +1,61 @@
...
+            self.payload = json.loads(self.settings.payload)
+
+    @make_block("Analysing trigger payload")
+    def execute(self):
+        if self.settings.event == "check_suite" and (self.payload["action"] in ["requested", "rerequested"]):
+            url = self.payload["repository"]["url"] + "/check-runs"
+            data = {"name": "CI tests", "head_sha": self.payload["check_suite"]["head_sha"]}
+            headers = {'Authorization': f"token {self.get_token(self.payload['installation']['id'])}",
+                       'Accept': 'application/vnd.github.antiope-preview+json'}
+            response = requests.post(url=url, json=data, headers=headers)
```
i-keliukh, line 40: Isn't it supposed to be easier to do this stuff using pygithub rather than POST-requests?

universum/github_handler.py
```diff
@@ -0,0 +1,61 @@
...
+            response.raise_for_status()
+
+        elif self.settings.event == "check_run" and \
+                (self.payload["action"] in ["requested", "rerequested", "created"]) and \
+                (str(self.payload["check_run"]["app"]["id"]) == str(self.settings.integration_id)):
+            self.trigger_build({
+                "GIT_REFSPEC": self.payload["check_run"]["check_suite"]["head_branch"],
```
i-keliukh, line 47: The loading of the JSON and using of the fields is placed to isolated parts of the code.
However, they are logically connected, so I prefer them to be inside the `execute` function.
Another issue: it will be hard to debug the issue with the format based on logs with this approach.
Probably, the parsing must be more resistant to the malformed data with proper error handling.

k-dovgan:
> Do you plan to implement tests for this functionality?

Yes I think of it, but not in this PR

universum/modules/vcs/github_vcs.py
k-dovgan: From the old token generation script. Also, I planned on adding links to GitHub documentation later, in docs-update PR

k-dovgan: > 
> 
> What does it refer to?
> Where is that 'general'?

GitHub user -> Settings -> Developer's settings -> App -> Edit
Will be explained in details later in docs-update PR

universum/modules/vcs/github_vcs.py
k-dovgan: Not sure at all what I meant back then when writing this script. Anyway, will be changed when updating documentation. Should be retrieved from webhook payload.

universum/modules/automation_server/jenkins_server.py
```diff
@@ -29,13 +34,20 @@ def __init__(self, *args, **kwargs):
...
+        if self.params:
+            processed_url += '?' + '&'.join(self.params)
 
-        self.out.log("Triggering url %s" % processed_url)
+        self.out.log(f"Triggering url {processed_url}")
         try:
-            requests.get(processed_url)
+            response = requests.post(processed_url)
+            response.raise_for_status()
+            self.out.log("Sucessfully triggered")
```
i-keliukh, line 50: Su**cc**essfully

universum/modules/automation_server/jenkins_server.py
```diff
@@ -29,13 +34,20 @@ def __init__(self, *args, **kwargs):
...
+        self.params = unify_argument_list(self.settings.param_list)
 
-    def trigger_build(self, revision):
-        processed_url = self.settings.trigger_url % revision
+    def trigger_build(self, param_dict=None):
+        processed_url = self.settings.trigger_url
+        if param_dict:
+            self.params.append(urllib.parse.urlencode(param_dict))
+        if self.params:
+            processed_url += '?' + '&'.join(self.params)
```
i-keliukh, line 44: Please use url handling library here

universum/github_handler.py
k-dovgan: Sure, this is just a blob

universum/github_handler.py
k-dovgan: Good question, I was sure I've researched it, but it doesn't seem right. I'll double check.

universum/modules/automation_server/jenkins_server.py
```diff
@@ -29,13 +34,20 @@ def __init__(self, *args, **kwargs):
...
 
-        self.out.log("Triggering url %s" % processed_url)
+        self.out.log(f"Triggering url {processed_url}")
         try:
-            requests.get(processed_url)
+            response = requests.post(processed_url)
```
i-keliukh, line 48: It is unusual to use POST http method with query string.
The traditional way is to use query string for GET and payload for POST.

universum/modules/vcs/github_vcs.py
i-keliukh: It is ok to remove help temporarily.
Please note, there is no link to 'general' under App -> Edit.

universum/modules/automation_server/jenkins_server.py
k-dovgan: https://plugins.jenkins.io/build-token-root/
> Just issue an Http GET or POST
Though I can switch to GET, no problem

k-dovgan:
> Also I noticed that code report is required to post result of checks to github.
> Probably, the same issue is valid with other code review systems.

Sure, why is this an issue?

universum/modules/vcs/github_vcs.py
```diff
@@ -17,7 +21,34 @@ def get_time():
...
+                            help="GitHub application ID (see 'general')")
+        parser.add_argument("--github-installation", "-ghs", dest="installation_id", metavar="GITHUB_INSTALLATION",
+                            help="in-project installation ID (see 'integrations & services')"
+                                 " - not needed for 'github-handler'")
+        parser.add_argument("--github-private-key", "-ghk", dest="key_path", metavar="GITHUB_PRIVATE_KEY",
+                            help="Application private key file path")
+
+    def __init__(self, *args, **kwargs):
+        super(GithubToken, self).__init__(*args, **kwargs)
+        with open(self.settings.key_path) as f:
```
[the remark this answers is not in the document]
k-dovgan: > Probably this is an issue in pygithub, but it still has to be investigated.

Got that one too, thought it was my own problem. Has to do something with `cryptography` module. Should I add it to dependencies for GitHub Handler?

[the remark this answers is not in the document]
i-keliukh: The reply is to different comment.
Anyway, if adding to dependencies fixes the issue, then add it.

i-keliukh:
> 
> 
> > Also I noticed that code report is required to post result of checks to github.
> > Probably, the same issue is valid with other code review systems.
> 
> Sure, why is this an issue?

Because code report is only providing annotations. But reporting checks success or failure must not depend on code report.

universum/modules/automation_server/jenkins_server.py
i-keliukh: This is just convention, not a rule

i-keliukh:
> 
> 
> > > Also I noticed that code report is required to post result of checks to github.
> > > Probably, the same issue is valid with other code review systems.
> > 
> > 
> > Sure, why is this an issue?
> 
> Because code report is only providing annotations. But reporting checks success or failure must not depend on code report.

Actually, it was just missing command-line parameter. Please ignore the comment.

universum/modules/vcs/github_vcs.py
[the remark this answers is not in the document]
k-dovgan: > With empty key_path or missing file this line crashes the app

Added TODO for further reference

universum/github_handler.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @staticmethod
+    def define_arguments(argument_parser):
+        argument_parser.add_argument('--event', '-e', dest='event', metavar="GITHUB_EVENT",
+                                     help='Currently parsed from "x-github-event" header')
+        argument_parser.add_argument('--payload', '-p', dest='payload', metavar="PAYLOAD",
+                                     help='<actual help coming later> leave "-" to read from stdin')
+
+    def __init__(self, *args, **kwargs):
+        super(GithubHandler, self).__init__(*args, **kwargs)
```
i-keliukh, line 24: This constructor is not needed

k-dovgan: explained in TODO

universum/github_handler.py
k-dovgan: Checked again. We're using this implementation:
https://github.com/PyGithub/PyGithub/blob/925764add54f557bfa446525a76ccb86180b577a/github/MainClass.py#L783
that doesn't provide any more functionality than already used. So, I guess, no? There's no easier way?

i-keliukh: Why don't we use another class then?

k-dovgan: Because we don't know of any other applicable classes.

universum/modules/vcs/github_vcs.py
```diff
@@ -17,7 +19,50 @@ def get_time():
...
+        parser.add_argument("--github-installation-id", "-gti", dest="installation_id",
+                            metavar="GITHUB_INSTALLATION_ID",
+                            help="Calculated out of webhook payload (real help coming soon)")
+
+    def __init__(self, *args, **kwargs):
+        super(GithubTokenWithInstallation, self).__init__(*args, **kwargs)
+        # TODO: add check for parameter
+
+    def get_token(self, installation_id=None):
+        super(GithubTokenWithInstallation, self).get_token(installation_id=self.settings.installation_id)
```
i-keliukh, line 62: Missing return

universum/modules/vcs/github_vcs.py
```diff
@@ -17,7 +19,50 @@ def get_time():
...
+    def define_arguments(argument_parser):
+        parser = argument_parser.get_or_create_group("GitHub", "GitHub repository settings")
+
+        parser.add_argument("--github-app-id", "-gta", dest="integration_id", metavar="GITHUB_APP_ID",
+                            help="GitHub application ID (real help coming soon)")
+        parser.add_argument("--github-private-key", "-gtk", dest="key_path", metavar="GITHUB_PRIVATE_KEY",
+                            help="Application private key file path")
+
+    def __init__(self, *args, **kwargs):
+        super(GithubToken, self).__init__(*args, **kwargs)
```
i-keliukh, line 34: Please use super().__init__(*args, **kwargs) instead.

universum/modules/automation_server/jenkins_server.py
```diff
@@ -18,24 +20,23 @@ class JenkinsServerForTrigger(BaseServerForTrigger):
     @staticmethod
     def define_arguments(argument_parser):
         parser = argument_parser.get_or_create_group("Jenkins variables", "Jenkins-specific parameters")
-        parser.add_argument('--jenkins-trigger-url', '-jtu', dest='trigger_url',
-                            help='Url to trigger, must include exactly one conversion specifier (%%s) to be '
-                                 'replaced by CL number, for example: http://localhost/%%s', metavar="URL")
+        parser.add_argument('--jenkins-trigger-url', '-jtu', dest='trigger_url', metavar="TRIGGER_URL",
+                            help='Url to trigger, with predefined parameters if any (e.g. token)')
```
i-keliukh, line 24: Somehow we lost %s from the help string.

universum/github_handler.py
```diff
@@ -0,0 +1,73 @@
...
+
+@needs_structure
+@needs_output
+class GithubHandler(GithubToken):
+
+    @staticmethod
+    def define_arguments(argument_parser):
+        argument_parser.add_argument('--event', '-e', dest='event', metavar="GITHUB_EVENT",
+                                     help='Currently parsed from "x-github-event" header')
+        argument_parser.add_argument('--payload', '-p', dest='payload', metavar="PAYLOAD",
```
i-keliukh, line 21: Attempt to pass -p results in the following error
```
__main__.py: error: ambiguous option: -p could match -pr, -p4c, -p4h, -p4s, -p4d, -p4m, -p4p, -p4u, -p4P
```

universum/github_handler.py
```diff
@@ -0,0 +1,73 @@
...
+                                     help='<actual help coming later> leave "-" to read from stdin')
+        argument_parser.add_argument('--trigger-url', '-t', dest='trigger_url', metavar="TRIGGER_URL",
```
i-keliukh, line 23: Another ambiguous option

universum/github_handler.py
```diff
@@ -0,0 +1,73 @@
...
+                                     help='<actual help coming later> including parameters like token')
+
+    def __init__(self, *args, **kwargs):
+        # TODO: add checks for params; add tests
+        super().__init__(*args, **kwargs)
+
+    @make_block("Analysing trigger payload")
+    def execute(self):
+        # TODO: refactor to real curl-style variable with '@' adn '-' support
+        # TODO: add some proper exception handling on payload contents; add tests
```
i-keliukh, line 33: Please add #TODO to handle http errors

PR #454, opened 2020-06-12
fix(github_handler): do not pass params not expected any more

k-dovgan:


universum/github_handler.py
```diff
@@ -92,9 +92,7 @@ def execute(self):
                 "GIT_CHECKOUT_ID": payload["check_run"]["head_sha"],
                 "GITHUB_CHECK_ID": payload["check_run"]["id"],
                 "GIT_REPO": payload["repository"]["clone_url"],
-                "GITHUB_APP_ID": self.settings.integration_id,
-                "GITHUB_INSTALLATION_ID": payload['installation']['id'],
-                "GITHUB_PRIVATE_KEY": self.settings.key_path
```
k-dovgan, line 97: this one is causing unexpected failures due to being renamed and removed

PR #456, opened 2020-06-15
fix(tests): mock github token

k-dovgan:


tests/test_report.py
```diff
@@ -14,26 +15,43 @@ def __init__(self, directory, client):
...
 
-        self.path = "http://localhost/repos" + \
-                    six.text_type(client.root_directory).rsplit("client", 1)[0] + \
-                    "server/check-runs/123"
+        repo_name = str(client.root_directory).rsplit("client", 1)[0]
+        self.path = "http://localhost/repos" + repo_name + "server/check-runs/123"
+
+
+@pytest.fixture()
+def mock_github_token(request):
```
i-keliukh, line 33: Please try to use monkeypatch in the test instead of implementing fixture on your own.
I am not sure whether it will look better, so we need to see how it looks.

tests/test_report.py
```diff
@@ -14,26 +14,30 @@ def __init__(self, directory, client):
...
     yield ReportEnvironment(tmpdir, git_client)
 
 
-@pytest.mark.xfail  # Until propper mocking implemented
-def test_github_run(http_check, report_environment):
+def test_github_run(http_check, report_environment, monkeypatch):
+    def mock_tocken(*args, **kwargs):
+        return "this is token"
+
+    monkeypatch.setattr(GithubToken, 'get_token', mock_tocken)
```
i-keliukh, line 40: tocken -> token

i-keliukh: Also, IMO, there is no need for named function

PR #457, opened 2020-06-17
feat(args): add multiline variable type

k-dovgan:
Also fix bug with 'metavar' and non-existing help in ModuleArgumentParser and deprecated 'iteritems' in github_vcs

### DIFF
--- a/tests/test_multiline_variables.py
+++ b/tests/test_multiline_variables.py
@@ -1,22 +1,34 @@
-from universum.lib.module_arguments import ModuleArgumentParser, multiline_argument
+import subprocess
 
 
-def test_multiline_vars(tmp_path, monkeypatch):
+script = """
+from universum.lib.module_arguments import ModuleArgumentParser
+from universum.lib.utils import read_multiline_option
+
+parser = ModuleArgumentParser()
+# parser.add_argument('--argument', '-a', metavar='ARGUMENT')
+parser.add_argument('--argument', '-a')
+
+namespace = parser.parse_args()
+print(read_multiline_option(namespace.argument))
+"""
+
+
+def test_success_multiline_variable(tmp_path):
     text = "This is text\nwith some line breaks\n"
-    parser = ModuleArgumentParser()
-    parser.add_argument("--argument", type=multiline_argument)
 
-    # Usual value
-    namespace = parser.parse_args(["--argument", text])
-    assert namespace.argument == text
+    # Direct input
+    script_path = tmp_path / "script.py"
+    script_path.write_text(script)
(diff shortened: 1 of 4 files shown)


tests/test_multiline_variables.py
```diff
@@ -0,0 +1,22 @@
...
+    assert namespace.argument == text
+
+    # File
+    tmp_file = tmp_path / "myfile.txt"
+    tmp_file.write_text(text)
+    namespace = parser.parse_args(["--argument", '@' + str(tmp_file)])
+    assert namespace.argument == text
+
+    # Stdin
+    monkeypatch.setattr('sys.stdin.read', lambda: text)
```
i-keliukh, line 20: As an alternative, you can spawn separate interpreter

tests/test_multiline_variables.py
```diff
@@ -0,0 +1,22 @@
...
+    text = "This is text\nwith some line breaks\n"
+    parser = ModuleArgumentParser()
+    parser.add_argument("--argument", type=multiline_argument)
+
+    # Usual value
+    namespace = parser.parse_args(["--argument", text])
+    assert namespace.argument == text
+
+    # File
+    tmp_file = tmp_path / "myfile.txt"
```
i-keliukh, line 14: I think, you can use tmpfile directly

universum/lib/module_arguments.py
```diff
@@ -5,6 +5,15 @@
 import sys
 
 
+def multiline_argument(sting):
```
i-keliukh, line 8: sting -> string

tests/test_multiline_variables.py
k-dovgan: I tried subprocess & multiprocessing, but liked this one more for better readability.

tests/test_multiline_variables.py
k-dovgan: pytest doesn't have a 'tmp_file' fixture, only 'tmp_path' and 'tmp_dir', which are both directories. I can switch to built-in `tempfile.NamedTemporaryFile` if needed, but I myself like pytest's fixtures better, and also we do use them in all other tests.

universum/lib/module_arguments.py
```diff
@@ -40,7 +39,10 @@ def _add_action(self, action):
                 action.help += "\n\nEnvironment variable: ${}".format(action.metavar)
             else:
                 if not isinstance(action, argparse._SubParsersAction):
-                    action.help += " [env: {}]".format(action.metavar)
+                    try:
+                        action.help += f" [env: {action.metavar}]"
+                    except TypeError:
+                        action.help = f"Also available as [env: {action.metavar}]"
```
i-keliukh, line 45: Please also fix sphinx case (line 39)

PR #458, opened 2020-06-19
feat(args): use newly added multiline variable reader

k-dovgan:


k-dovgan:
As an alternative 'read_multiline_variable' can be united with 'check_required_option' and accept settings as parameter and raise 'empty var!' exception itself.

universum/github_handler.py
```diff
@@ -59,16 +52,26 @@ def __init__(self, *args, **kwargs):
                     Please specify this parameter by using '--trigger-url' ('-tu')
                     command line parameter or by setting TRIGGER_URL environment variable.
                 """)
+        try:
+            self.payload = utils.read_multiline_option(self.settings.payload)  # actually may throw AttributeError
+            if not self.payload:
+                raise AttributeError("Empty payload")
+        except AttributeError:
+            raise IncorrectParameterError("""
```
i-keliukh, line 60: cleandoc is missing

universum/github_handler.py
```diff
@@ -59,16 +52,26 @@ def __init__(self, *args, **kwargs):
                     Please specify this parameter by using '--trigger-url' ('-tu')
                     command line parameter or by setting TRIGGER_URL environment variable.
                 """)
+        try:
```
i-keliukh, line 55: This whole block should be moved to utils as well.

k-dovgan: that's exactly what I've written in comment above. The question is: should it become a part of 'read' function? Should the function become 'read_and_check' or two separate functions?

universum/lib/utils.py
```diff
@@ -93,21 +93,21 @@ def check_required_option(settings, setting_name, error_message):
...
+                with open(value.lstrip('@')) as file_name:
+                    result = file_name.read()
+            except FileNotFoundError as e:
+                raise IncorrectParameterError(f"Error reading argument {setting_name} from file {e.filename}: no such file")
+        elif value == '-':
+            result = sys.stdin.read()
+        else:
+            result = value
+        if not result:
+            raise AttributeError('empty value')
```
i-keliukh, line 107: I think raising IncorrectParameterError has the benefit of keeping original stack trace and is cleaner.

PR #459, opened 2020-06-19
refactor(github_vcs): move non-mandatory import to __init__

k-dovgan:
Also specify vcs-related requirements in 'setup.py': package 'cryptography' has its own dependency on 'cffi', so installing 'cryptography' will also install 'cffi'. However, installing 'cffi' alone is not enough for correct module work.

### DIFF
--- a/doc/prerequisites.rst
+++ b/doc/prerequisites.rst
@@ -39,7 +39,8 @@ GitHub
 If used as GitHub Handler:
 
 * :mod:`pygithub` module for Python (use ``sudo pip install pygithub``)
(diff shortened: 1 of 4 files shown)


k-dovgan:
Got following errors in clear environment with `pygithub` installed:
```
KeyError: 'RS256'
NotImplementedError: Algorithm 'RS256' could not be found. Do you have cryptography installed?
```
So just `cffi` is not enough.

i-keliukh:
> So just `cffi` is not enough.

It is not clear how did you make this conclusion.

PR #461, opened 2020-06-24
feat(github-handler): add payload parsing error handling and tests

k-dovgan:


tests/test_github_handler.py
```diff
@@ -0,0 +1,142 @@
...
+        self.settings.GithubToken.key = "this is key"
+
+
+@pytest.fixture()
+def github_handler_environment(tmpdir):
+    yield GithubHandlerEnvironment(tmpdir)
+
+
+def test_success_github_handler(http_check, github_handler_environment, monkeypatch):
+    monkeypatch.setattr(GithubToken, 'get_token', lambda *args, **kwargs: "this is token")
```
i-keliukh, line 30: This line is repeated in every test.
Maybe put it into own autouse fixture?

tests/test_github_handler.py
```diff
@@ -0,0 +1,142 @@
...
+    url = "http://example.com/check-runs"
+    github_handler_environment.settings.GithubHandler.payload = """
+{
+  "action": "requested",
+  "repository": {
+    "url": "http://example.com"
```
i-keliukh, line 36: Code duplication

tests/test_github_handler.py
```diff
@@ -0,0 +1,142 @@
+# pylint: disable = redefined-outer-name, abstract-method
+
+import pytest
+
+from universum import __main__
```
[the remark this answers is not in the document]
k-dovgan: No. Maybe I'm doing this the wrong way, but when I used `import universum`, I got "there's no function called 'run' in package 'universum'" exception. Import universum only imports what is in __init__.py

tests/test_github_handler.py
```diff
@@ -0,0 +1,142 @@
...
+  },
+  "check_suite": {
+    "head_sha": "1234"
+  },
+  "installation": {
+    "id": "1234"
+  }
+}
+"""
+    http_check.assert_success_and_collect(__main__.run, github_handler_environment.settings, url=url, method="POST")
```
i-keliukh, line 46: Assuming the main functionality of the handler for check_suite event is parsing of certain fields from the payload, it would make sense to check the correctness of the parsing within the test.
Specifically, the following fields are not checked directly:
 - payload["action"]
 - payload["check_suite"]["head_sha"]
 - payload['installation']['id']

You already indirectly check payload["repository"]["url"], which is fine.

AFAIU, payload["action"] should be checked by other tests: different value - not trigger.
Also, checking of payload['installation']['id'] requires more smart mock implementation, which shouldn't be a problem.

tests/test_github_handler.py
```diff
@@ -0,0 +1,142 @@
...
+      "check_suite": {
+        "head_sha": "1234"
+      }
+    }
+    """
+    assert __main__.run(github_handler_environment.settings)
+    stdout_checker.assert_has_calls_with_param("Could not find key 'installation' in provided payload")
+
+
+def test_error_github_handler_no_github_server(stdout_checker, github_handler_environment, monkeypatch):
```
i-keliukh, line 94: It is bad idea to test against real server

tests/test_github_handler.py
```diff
@@ -0,0 +1,142 @@
...
+  "installation": {
+    "id": "1234"
+  }
+}
+"""
+    assert __main__.run(github_handler_environment.settings)
+    stdout_checker.assert_has_calls_with_param("404 Client Error: Not Found for url: http://example.com/check-runs")
+
+
+def test_error_github_handler_no_jenkins_server(stdout_checker, github_handler_environment, monkeypatch):
```
i-keliukh, line 116: We also need positive check_run test with checking parsing of parameters

universum/github_handler.py
```diff
@@ -63,44 +60,50 @@ def __init__(self, *args, **kwargs):
...
+                response.raise_for_status()
+                self.out.log("Successfully triggered")
+
+            else:
+                self.out.log("Unhandled event, skipping...")
+        except KeyError as error:
+            raise CriticalCiException(f"Could not find key {error} in provided payload:\n{payload}")
+        except TypeError:
+            raise CriticalCiException("Parsed payload JSON does not correspond to expected format")
+        except (requests.HTTPError, requests.ConnectionError, ValueError) as error:
```
i-keliukh, line 105: Each of these exceptions must be tested

tests/test_github_handler.py
```diff
@@ -15,67 +15,99 @@ def __init__(self, directory):
...
+def mock_token(monkeypatch):
     monkeypatch.setattr(GithubToken, 'get_token', lambda *args, **kwargs: "this is token")
+
+
+def test_success_github_handler_check_suite(http_check, github_handler_environment):
+    github_handler_environment.settings.GithubHandler.event = "check_suite"
+    github_handler_environment.settings.GithubHandler.payload = github_handler_environment.check_suite_payload
+    http_check.assert_success_and_collect(__main__.run, github_handler_environment.settings,
+                                          url=github_handler_environment.check_suite_url, method="POST")
+    http_check.assert_request_headers_contained('Accept', 'application/vnd.github.antiope-preview+json')
```
i-keliukh, line 75: In my opinion, this is change detector test.
With this test we have duplication of this magic string.
However, since this is what github.com demands, if its api changes, we will have failure in our tests and in production.

i-keliukh:
The name of the PR is misleading.

universum/github_handler.py
k-dovgan: replaced with TODO until further implementation

PR #466, opened 2020-07-07
fix: remove shebangs and absolute imports

### LINKED ISSUE #214: Run static analyzers as file
Originally created on Tue, 20 Nov 2018 15:49:44 +0200

When you try to run analyzers as file you have import error
Example:
Run command `./analyzers/uncrustify.py -h` and get an error message:
```
Traceback (most recent call last):
  File "analyzers/uncrustify.py", line 12, in <module>
    from . import utils
ValueError: Attempted relative import in non-package
```

k-dovgan:
Resolves #214 as no longer applicable

We remove shebangs as Universum is no longer recommended to be run as a script (as in `./universum/__main__.py`).
Recommended way instead is running it via `python3.7 -m universum`, to fix the entry point issues (such as not being able to pass exit codes), and also to make used python version clear to user.

We replace absolute imports with relative as they were only needed for script launching to work, and caused a very non-obvious scenario for developing, when a main module launched locally imported installed version of `universum` folder instead of local (and presumably changed) one.

### DIFF
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -220,14 +220,13 @@ class UniversumRunner:
             f.write(config)
         return file_path
 
-    def run(self, config: str, vcs_type: str = "none", force_installed=False, workdir=None,
-            additional_parameters="", environment=None, expected_to_fail=False):
+    def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
+            additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
 
-        # We can only collect coverage from local sources
         if force_installed:
             cmd = "python3.7 -I -m universum"
         elif utils.is_pycharm() or workdir:
-            # Changing workdir will also lead to using installed Universum instead of local sources
+            # Changing workdir may also lead to using installed Universum instead of local sources
             cmd = "python3.7 -m universum"
         else:
             cmd = f"coverage run --branch --append --source='{self.working_dir}' -m universum"
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -223,10 +223,16 @@ class UniversumRunner:
     def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
             additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
 
-        # workdir is ignored unless force_installed;
-        # if installed to make sure local sources are not used by default the directory is changed
+        # as nonci changes workdir to project root, Universum is always an external moudule
+        # therefore for nonci it should always be installed to system
+        if self.nonci:
+            force_installed = True
+
+        # workdir is ignored unless force_installed is True
         if force_installed:
             if not workdir:
+                # when run from directory containing Universum sources, installed module will not be used
+                # therefore directory should be changed to any other to make sure 'python -m universum' still works
                 workdir = os.path.join(self.working_dir, "another_directory")
                 os.makedirs(workdir, exist_ok=True)
         else:
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -223,15 +223,20 @@ class UniversumRunner:
     def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
             additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
 
-        if utils.is_pycharm():
-            cmd = "{0}/universum/__main__.py".format(self.working_dir)
-        else:
-            cmd = "coverage run --branch --append --source='{0}' '{0}/universum/__main__.py'" \
-                .format(self.working_dir)
-
-        # We cannot collect coverage from installed module, so we run it only if specifically told so
+        # workdir is ignored unless force_installed;
+        # if installed to make sure local sources are not used by default the directory is changed
         if force_installed:
+            if not workdir:
+                workdir = os.path.join(self.working_dir, "another_directory")
+                os.makedirs(workdir, exist_ok=True)
+        else:
+            workdir = self.working_dir
+
+        # We cannot collect coverage from installed module
+        if utils.is_pycharm() or force_installed:
             cmd = "python3.7 -m universum"
+        else:
+            cmd = f"coverage run --branch --append --source='.' 'python3.7 -m universum'"
 
         if self.nonci:
             cmd += ' nonci'
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -222,11 +222,17 @@ class UniversumRunner:
 
     def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
             additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
+        """
+        `force_installed` launches python with '-I' option, that ensures the non-installed universum sources
+        will not be used instead of those installed into system. Without '-I' option `python3.7 -m` will first
+        try to launch universum from sources in `workdir` if there are any. That is why, if `workdir` is not
+        default and there are no universum sources in specified `workdir`, the preinstalled universum will
+        be ran as in case of `force_installed`.
+        """
 
         if force_installed:
             cmd = "python3.7 -I -m universum"
         elif utils.is_pycharm() or workdir:
-            # Changing workdir may also lead to using installed Universum instead of local sources
             cmd = "python3.7 -m universum"
         else:
             cmd = f"coverage run --branch --append --source='{self.working_dir}' -m universum"
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -236,7 +236,7 @@ class UniversumRunner:
         if utils.is_pycharm() or force_installed:
             cmd = "python3.7 -m universum"
         else:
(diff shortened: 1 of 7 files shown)


tests/deployment_utils.py
```diff
@@ -223,15 +223,20 @@ def _create_temp_config(self, config: str):
     def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
             additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
 
-        if utils.is_pycharm():
-            cmd = "{0}/universum/__main__.py".format(self.working_dir)
+        # workdir is ignored unless force_installed;
```
i-keliukh, line 226: unless force_installed is True?

tests/deployment_utils.py
```diff
@@ -223,15 +223,20 @@ def _create_temp_config(self, config: str):
...
+        # if installed to make sure local sources are not used by default the directory is changed
+        if force_installed:
+            if not workdir:
+                workdir = os.path.join(self.working_dir, "another_directory")
```
i-keliukh, line 230: I don't understand where this"another_directory" is coming from and what is the goal.
We already have quite dirty working directory after running tests.

tests/test_integration.py
```diff
@@ -77,7 +77,7 @@ def test_background_steps(docker_main_and_nonci):
 configs = Variations([dict(name="Bad step", command=["ls", "not_a_file"]),
                       dict(name="Good bg step", command=["touch", "file"],
                       background=True, artifacts="file")])
-""")
+""", force_installed=docker_main_and_nonci.nonci)
```
i-keliukh, line 80: You are passing the value you've got from the runner to the runner.
I guess, the runner can do that by itself.

tests/test_nonci.py
```diff
@@ -29,7 +30,7 @@ def test_launcher_output(docker_nonci):
         f"bash -c 'mkdir {cwd}/artifacts; echo \"Old artifact\" > {cwd}/artifacts/test_nonci.txt'")
 
     docker_nonci.project_root = None
-    console_out_log = docker_nonci.run(config.format(cwd), workdir=cwd)
+    console_out_log = docker_nonci.run(config.format(cwd), workdir=cwd, force_installed=True)
```
i-keliukh, line 33: The reason of these changes is not clear from the PR.
I guess you can write it into the commit message, but as of now I don't understand it.

tests/deployment_utils.py
k-dovgan: The thing is, `python -m module_name` is a little bit dirty itself. If module is installed in system, this command will work from any place on computer; if module _is not installed,_ it will work only from the directory where module's sources are located.
So if we want to make sure Universum was installed correctly, we _have_ to launch it from anywhere else than workdir.

i-keliukh: Then please update the comment to make the intention more clear.

tests/test_integration.py
k-dovgan: I'm not sure how exactly runner should do it itself. Do you suggest to always require `force_installed` in case of nonci?

i-keliukh: Probably. To answer I have to understand the intention of this change.
My comment is just observation.

tests/test_nonci.py
k-dovgan: Updated the description of the PR.

i-keliukh: Ok, the PR description looks nice, but the intention of this exact line (and similar ones) is still not clear.

tests/deployment_utils.py
```diff
@@ -223,15 +223,26 @@ def _create_temp_config(self, config: str):
     def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
             additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
 
-        if utils.is_pycharm():
-            cmd = "{0}/universum/__main__.py".format(self.working_dir)
-        else:
-            cmd = "coverage run --branch --append --source='{0}' '{0}/universum/__main__.py'" \
-                .format(self.working_dir)
+        # as nonci changes workdir to project root, Universum is always an external moudule
```
i-keliukh, line 226: moudule -> module

tests/deployment_utils.py
```diff
@@ -223,10 +223,16 @@ def _create_temp_config(self, config: str):
     def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
             additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
 
-        # workdir is ignored unless force_installed;
-        # if installed to make sure local sources are not used by default the directory is changed
+        # as nonci changes workdir to project root, Universum is always an external moudule
```
i-keliukh, line 226: The connection these two statements is not clear to me:
1. nonci changes the workdir to project root
2. Universum is an external module

nonci is one part of the Universum. So, the respective nonci implementation is chosen depending on how the Universum is launched. The change of the workdir happens only after that.

Probably, I just misunderstand the comment.

tests/deployment_utils.py
```diff
@@ -220,18 +220,14 @@ def _create_temp_config(self, config: str):
...
-
-        # We cannot collect coverage from installed module, so we run it only if specifically told so
-        if force_installed:
+        # We cannot collect coverage from installed module
+        if utils.is_pycharm() or workdir:
             cmd = "python3.7 -m universum"
+        else:
+            cmd = f"coverage run --branch --append --source='{self.working_dir}' -m universum"
 
         if self.nonci:
```
k-dovgan, line 232: This if block is the reason we cannot check the artifacts for nonci. Because passing `-ad` in in _basic_args, that are not added to nonci; and without correct artifact directory the artifacts are not retrieved from docker.
Should be fixed in another review

tests/deployment_utils.py
```diff
@@ -220,18 +220,14 @@ def _create_temp_config(self, config: str):
...
-        if utils.is_pycharm():
-            cmd = "{0}/universum/__main__.py".format(self.working_dir)
-        else:
-            cmd = "coverage run --branch --append --source='{0}' '{0}/universum/__main__.py'" \
-                .format(self.working_dir)
-
-        # We cannot collect coverage from installed module, so we run it only if specifically told so
-        if force_installed:
+        # We cannot collect coverage from installed module
+        if utils.is_pycharm() or workdir:
```
i-keliukh, line 227: With this change the `workdir` is only used for launching the installed version.
While this unification is generally ok, it should be explicitly expressed in the code.

tests/test_deployment.py
```diff
@@ -6,45 +11,66 @@
...
 
 def test_minimal_install(clean_docker_main):
+    # Create tmpdir in container to make sure local universum sources are not used instead of installed
+    tmpdir = os.path.join(clean_docker_main.working_dir, randomize_name('tmpdir'))
+    clean_docker_main.environment.assert_successful_execution(f"mkdir '{tmpdir}'")
+
     # Run without parameters
-    log = clean_docker_main.environment.assert_unsuccessful_execution("universum")
-    assert "command not found" not in log
+    log = clean_docker_main.environment.assert_unsuccessful_execution("python3.7 -m universum")
```
i-keliukh, line 19: This block and the cleanup block should be changed to the fixture.
Probably they can be merged into clean_docker fixture.

tests/deployment_utils.py
```diff
@@ -220,11 +220,14 @@ def _create_temp_config(self, config: str):
...
+    def run(self, config: str, vcs_type: str = "none", force_installed=False, workdir=None,
             additional_parameters="", environment=None, expected_to_fail=False):
 
-        # We cannot collect coverage from installed module
-        if utils.is_pycharm() or workdir:
+        # We can only collect coverage from local sources
+        if force_installed:
+            cmd = "python3.7 -I -m universum"
+        elif utils.is_pycharm() or workdir:
+            # Changing workdir will also lead to using installed Universum instead of local sources
```
i-keliukh, line 230: Changing workdir may also lead to using installed Universum instead of local sources.

tests/deployment_utils.py
```diff
@@ -220,11 +220,14 @@ def _create_temp_config(self, config: str):
             f.write(config)
         return file_path
 
-    def run(self, config: str, vcs_type: str = "none", workdir=None,
+    def run(self, config: str, vcs_type: str = "none", force_installed=False, workdir=None,
             additional_parameters="", environment=None, expected_to_fail=False):
 
-        # We cannot collect coverage from installed module
-        if utils.is_pycharm() or workdir:
+        # We can only collect coverage from local sources
```
i-keliukh, line 226: The comment is now outdated

tests/deployment_utils.py
```diff
@@ -220,18 +220,17 @@ def _create_temp_config(self, config: str):
             f.write(config)
         return file_path
 
-    def run(self, config: str, force_installed: bool = False, vcs_type: str = "none",
-            additional_parameters="", environment=None, expected_to_fail=False, workdir=None):
+    def run(self, config: str, vcs_type: str = "none", force_installed=False, workdir=None,
```
i-keliukh, line 223: Could you, please, change the order of parameters back?
This will minimize the amount of changes on diff.

Issue #361, opened 2019-10-24
Interactive terminal output looks awful in logs


k-dovgan:
Module sh opens terminal for each Universum step and logs all stdout from it; in case of interactive changes (e.g. status bars) these logs are very long and hardly changed.

PR #468, opened 2020-07-09
fix(configs): remove pip progress bars (requires pip>=10)

k-dovgan:
Done in scope of #361; though detecting and processing by Universum interactive progress bars doesn't actually seem possible at the moment


k-dovgan:
> What about pip command line used in tests?

1. Pip modules are installed in docker containers once before all tests
2. The output of these commands is hidden unless something went wrong
3. Actually, we do not update Pip to latest version in docker container, it has pip 9.0.1; so the '--progress-bar off' is not even applicable. And taking in mind it is not showed anyway, I don't think it's worth adding?

Issue #218, opened 2019-01-29
Set environment variable 'CODE_REPORT_FILE' for code analyzers in Universum


k-dovgan:
Originally created on Fri, 30 Nov 2018 14:48:34 +0200

Due to task 'Add support of code analyzers CODE_REPORT_FILE as environment variable'

PR #469, opened 2020-07-10
fix(code_report): replace code report pseudo-variable not only in command

### LINKED ISSUE #286: The ${CODE_REPORT_FILE} variable is not being replaced
## Description

There is no replacement of the pseudo-variable "${CODE_REPORT_FILE}" to the file name (generated by Universum). The "universum_pylint" module is used via specific way: through "bash -c ..." call. The `"code_report=True" `is added

Full config:
```
import os
from distutils.spawn import find_executable

from _universum.configuration_support import Variations, get_project_root

# Unittests and pylint check #
check = Variations([
    dict(name="Run",
         critical=True)
])
configs = check * Variations([
    dict(name=" python3.5 symlink create",
         command=["bash", "-c", "ln -s {} python3".format(find_executable('python3.5'))]),
    dict(name=" pylint3",
         code_report=True,
         command=["bash", "-c",
                  "export PATH=%s:$PATH; "
                  "universum_pylint --python-version 3 "
                  "--result-file ${CODE_REPORT_FILE} "
                  "--files *.py " % get_project_root()]),

])
```

## Steps to reproduce

 * Prepare the config file as mentioned above
 * Run the command: 
`universum --clean-build --vcs-type none -lo console --file-source-dir ./test_project/ --launcher-config-path ./universum_config.py`

## Expected result

All steps are executed without errors

## Actual result

The ${CODE_REPORT_FILE} variable is not changed to the real file name during config parsing by Universum. The bash interpreter try to execute `universum_pylint --result-file ${CODE_REPORT_FILE} --files *.py` command but the "CODE_REPORT_FILE" variable is empty. That is why the command is interpreted in `universum_pylint --result-file --files *.py` and error is occurred

Log:
```
5. Executing build steps
 |   5.1.  [  +  ] Run
 |      |   5.1.1.  [ 1/2 ] Run python3.5 symlink create
 |      |        |   $ /bin/bash -c ln -s /usr/bin/python3.5 python3
 |      |        |   
 |      |        └ [Success]
 |      |   
 |      |   5.1.2.  [ 2/2 ] Run pylint3
 |      |        |   $ /bin/bash -c export PATH=/home/test/temp:$PATH; universum_pylint --python-version 3 --result-file ${CODE_REPORT_FILE} --files *.py 
 |      |        |   stderr: usage: universum_pylint [-h] [--files FILE_LIST [FILE_LIST ...]]
 |      |        |   stderr:                         [--rcfile RCFILE] [--python-version {2,3}]
 |      |        |   stderr:                         [--result-file RESULT_FILE]
 |      |        |   stderr: universum_pylint: error: argument --result-file: expected one argument
 |      |        |   Error: Module sh got exit code 2
 |      |        |   
 |      |        └ [Failed]
 |      |   
 |      └ [Success]
 |   
 |   Critical step failed. All further configurations will be skipped
 └ [Success]
```

## Environment

 - OS: Ubuntu 14.04
 - Python version 2.7.6
 - Universum version 0.17.0

k-dovgan:
Added iteration to not only replace in 'artifacts', but in 'report_artifacts' and generally anywhere applicable.

Resolves #286

### DIFF
--- a/tests/test_code_report.py
+++ b/tests/test_code_report.py
@@ -5,6 +5,9 @@ from typing import List
 
 import pytest
 
+from universum import __main__
+from . import utils
+
 
 @pytest.fixture(name='runner_with_pylint')
 def fixture_runner_with_pylint(docker_main):
@@ -79,16 +82,27 @@ def test_pylint_analyzer_wrong_params(runner_with_pylint, args, expected_log):
     assert expected_log in log
 
 
-def test_code_report_extended_arg_search(runner_with_pylint):
-    runner_with_pylint.local.root_directory.join("source_file.py").write(source_code + '\n')
+def test_code_report_extended_arg_search(tmpdir, stdout_checker):
+    env = utils.TestEnvironment(tmpdir, "main")
+    env.settings.Vcs.type = "none"
+    env.settings.LocalMainVcs.source_dir = str(tmpdir)
+    env.settings.Main.no_diff = True  # TODO: remove when issue #477 is fixed
+
+    tmpdir.join("source_file.py").write(source_code + '\n')
+
     config = """
 from universum.configuration_support import Variations
 
-configs = Variations([dict(name="Run static pylint", code_report=True, artifacts="${CODE_REPORT_FILE}", command=[
+configs = Variations([dict(name="Run static pylint", code_report=True, artifacts="${{CODE_REPORT_FILE}}", command=[
     'bash', '-c',
-    'python3.7 -m universum.analyzers.pylint --result-file=\"${CODE_REPORT_FILE}\" --python-version=3 --files source_file.py'
+    'cd {0} && python3.7 -m universum.analyzers.pylint --result-file=\"${{CODE_REPORT_FILE}}\" --python-version=3 \
+--files {1}/source_file.py'
 ])])"""
 
-    log = runner_with_pylint.run(config)
-    assert re.findall(log_fail, log)
-    assert os.path.exists(os.path.join(runner_with_pylint.artifact_dir, "Run_static_pylint.json"))
+    env.configs_file.write(config.format(os.getcwd(), str(tmpdir)))
+
+    res = __main__.run(env.settings)
+
+    assert res == 0
+    stdout_checker.assert_has_calls_with_param(log_fail, is_regexp=True)
+    assert os.path.exists(os.path.join(env.settings.ArtifactCollector.artifact_dir, "Run_static_pylint.json"))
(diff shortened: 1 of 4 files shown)


tests/test_code_report.py
```diff
@@ -76,3 +77,18 @@ def test_pylint_analyzer_wrong_params(runner_with_pylint, args, expected_log):
     log = runner_with_pylint.run(get_config(args))
     assert re.findall(r'Run static pylint - [^\n]*Failed', log)
     assert expected_log in log
+
+
+def test_code_report_extended_arg_search(runner_with_pylint):
```
i-keliukh, line 82: This test needs some comments.
The portion with expanding the "variable" in artifacts is ok, but with bash is it not clear enough.
BTW, why it didn't work with bash?

universum/modules/code_report_collector.py
```diff
@@ -39,13 +39,22 @@ def prepare_environment(self, project_configs):
...
-                if temp_filename in i:
-                    name = utils.calculate_file_absolute_path(self.report_path, item.get("name")) + ".json"
-                    actual_filename = os.path.join(self.report_path, name)
-                    item["command"][enum] = item["command"][enum].replace(temp_filename, actual_filename)
+            name = utils.calculate_file_absolute_path(self.report_path, item.get("name")) + ".json"
+            actual_filename = os.path.join(self.report_path, name)
+
+            for key in item:
+                if key == "command":
+                    for enum, i in enumerate(item[key]):
```
i-keliukh, line 48: Why we need enumerate at all?

i-keliukh:
Can we fix #218 as well?

tests/test_code_report.py
k-dovgan: Because this exact issue is outdated, and some older versions of code searched for fake variable in `cmd`, not in every `cmd` member. It was then fixed, but the issue was however not closed.

universum/modules/code_report_collector.py
k-dovgan: To address one exact list member when running replace in line 50, of course

tests/test_code_report.py
i-keliukh: So, if I remove the reference to the variable from artifacts key and revert the fix, would the test succeed?

universum/modules/code_report_collector.py
i-keliukh: The i` variable is used only in line 49.
I guess it should work even if you use `item[key][enum]` instead.

Also, I am not sure the `if` on line 49 is needed at all. Can you just always call `replace`?

tests/test_code_report.py
k-dovgan: Sure.

i-keliukh: Final question: when we fixed the issue with bash?

Issue #273, opened 2019-03-04
Implement static type checking using mypy


i-keliukh:
[mypy](http://mypy-lang.org/) provides mechanism for statically checking types. Applying static typing requires adding type annotations (comments) and some refactoring.

Specifically, the entire `gravity` module and `Dependency` class from it should be changed a little. The `Dependency` must inherit from `Generic[T]` and get actual type using `typing_inspect.get_origin()` and `typing_inspect.get_args()`. The type has to be propagated to other places as well.

PR #476, opened 2020-07-27
fix(structure_handler, git_vcs, vcs): pre-existing mypy errors

o-andrieiev:
Related to https://github.com/Samsung/Universum/issues/273

Without the fix, checking 'main' using mypy generates some errors. With the fix the only remaining errors are due to lack of static type annotations in the dependencies (third-party libraries).
To make static analysis of 'vcs' module meaningful, some minor refactoring was needed.
Also, added static type annotations highlighted some missing dependencies in 'git_vcs'.

!!! Please note that no tests have been run yet - this is preliminary pull request for some early feedback.

# Checklist:

- [ ✓ ] I have performed a self-review of my own code

### DIFF
--- a/universum/modules/structure_handler.py
+++ b/universum/modules/structure_handler.py
@@ -4,7 +4,7 @@ from .. import configuration_support
 from ..lib.ci_exception import SilentAbortException, StepException, CriticalCiException
 from ..lib.gravity import Module, Dependency
 from .output import needs_output
-
+from typing import List, Optional
 __all__ = [
     "needs_structure"
 ]
@@ -51,15 +51,15 @@ class Block:
     True
     """
 
-    def __init__(self, name: str, parent: 'Block' = None):
-        self.name = name
-        self.status = "Success"
-        self.children = []
+    def __init__(self, name: str, parent: Optional[Block] = None):
+        self.name: str = name
+        self.status: str = "Success"
+        self.children: List[Block] = []
 
-        self._parent = parent
-        self.number = ''
-        if self.parent:
-            self.parent.children.append(self)
+        self._parent: Optional[Block] = parent
+        self.number: str = ''
+        if parent:
+            parent.children.append(self)
             self.number = '{}{}.'.format(parent.number, len(parent.children))
 
     def __str__(self) -> str:
@@ -67,7 +67,7 @@ class Block:
         return '{} - {}'.format(result, self.status) if not self.children else result
 
     @property  # getter
-    def parent(self) -> 'Block':
+    def parent(self) -> Optional[Block]:
         return self._parent
 
     def is_successful(self) -> bool:
@@ -78,11 +78,10 @@ class Block:
 class StructureHandler(Module):
     def __init__(self, *args, **kwargs):
         super(StructureHandler, self).__init__(*args, **kwargs)
-        block_structure = Block("Universum")
-        self.current_block = block_structure
-        self.configs_current_number = 0
-        self.configs_total_count = 0
-        self.active_background_steps = []
+        self.current_block: Block = Block("Universum")
+        self.configs_current_number: int = 0
+        self.configs_total_count: int = 0
+        self.active_background_steps: List[Callable[[Any], Any]] = []
 
     def open_block(self, name):
         new_block = Block(name, self.current_block)
--- a/universum/modules/vcs/git_vcs.py
+++ b/universum/modules/vcs/git_vcs.py
@@ -2,7 +2,7 @@ import glob
 import importlib
 import os
 
-from .base_vcs import BaseVcs, BaseDownloadVcs
+from .base_vcs import BaseVcs, BaseDownloadVcs, BasePollVcs, BaseSubmitVcs
 from ..output import needs_output
 from ..structure_handler import needs_structure
 from ...lib import utils
@@ -196,7 +196,7 @@ class GitMainVcs(GitVcs, BaseDownloadVcs):
 
 
 @needs_output
-class GitSubmitVcs(GitVcs):
+class GitSubmitVcs(GitVcs, BaseSubmitVcs):
     @staticmethod
     def define_arguments(argument_parser):
         parser = argument_parser.get_or_create_group("Git")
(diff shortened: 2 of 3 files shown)


universum/modules/structure_handler.py
```diff
@@ -51,23 +51,23 @@ class Block:
...
+        self.status: str = "Success"
+        self.children: List[Block] = []
 
-        self._parent = parent
-        self.number = ''
-        if self.parent:
-            self.parent.children.append(self)
+        self._parent: Optional[Block] = parent
+        self.number: str = ''
+        if parent:
```
o-andrieiev, line 61: this code never executed before since 'self' doesn't have a 'parent' field - only '_parent'

universum/modules/structure_handler.py
```diff
@@ -78,11 +78,10 @@ def is_successful(self) -> bool:
 class StructureHandler(Module):
     def __init__(self, *args, **kwargs):
         super(StructureHandler, self).__init__(*args, **kwargs)
-        block_structure = Block("Universum")
```
o-andrieiev, line 81: I'm not sure what was the purpose of this variable

universum/modules/structure_handler.py
```diff
@@ -78,11 +78,10 @@ def is_successful(self) -> bool:
...
-        self.current_block = block_structure
-        self.configs_current_number = 0
-        self.configs_total_count = 0
-        self.active_background_steps = []
+        self.current_block: Block = Block("Universum")
+        self.configs_current_number: int = 0
+        self.configs_total_count: int = 0
+        self.active_background_steps: List[Callable[[Any], Any]] = []
```
o-andrieiev, line 84: we should really apply some restrictions here, but I don't know much about possible values in 'active_background_steps'

universum/modules/vcs/git_vcs.py
```diff
@@ -196,7 +196,7 @@ def copy_cl_files_and_revert(self):
 
 
 @needs_output
-class GitSubmitVcs(GitVcs):
+class GitSubmitVcs(GitVcs, BaseSubmitVcs):
```
o-andrieiev, line 199: missing dependency

universum/modules/vcs/git_vcs.py
```diff
@@ -304,7 +304,7 @@ def submit_new_change(self, description, file_list, review=False, edit_only=Fals
         return change
 
 
-class GitPollVcs(GitVcs):
+class GitPollVcs(GitVcs, BasePollVcs):
```
o-andrieiev, line 307: missing dependency

universum/modules/vcs/vcs.py
```diff
@@ -20,35 +22,43 @@
 ]
 
 
-def create_vcs(class_type=None):
+def create_vcs(class_type: str = None) -> Type[ProjectDirectory]:
+    driver_factory_class: Union[
```
o-andrieiev, line 26: this way all the drivers have to be of the same type - poll, submit or download

universum/modules/vcs/vcs.py
```diff
@@ -87,19 +97,9 @@ def __init__(self, *args, **kwargs):
...
-                    driver_factory = self.git_driver_factory
-                elif self.settings.type == "gerrit":
-                    driver_factory = self.gerrit_driver_factory
-                elif self.settings.type == "github":
-                    driver_factory = self.github_driver_factory
-                else:
-                    driver_factory = self.perforce_driver_factory
-            except AttributeError:
+                self.driver = driver_factory[self.settings.type]()
+            except AttributeError | KeyError:
```
o-andrieiev, line 101: I assume KeyError can never happen here due to option filtering. Should I remove this precaution?

k-dovgan: I'd say, yes.

universum/modules/structure_handler.py
k-dovgan: It was added to improve readability, but I think it can be removed now.

universum/modules/structure_handler.py
i-keliukh: No, there is a 'parent' property.

universum/modules/structure_handler.py
i-keliukh: It requires explicitly defining class for background step.
This is a simple refactoring and you can do it as a part of this PR.

i-keliukh: Callable and Any are not defined

universum/modules/structure_handler.py
```diff
@@ -51,23 +51,23 @@ class Block:
     True
     """
 
-    def __init__(self, name: str, parent: 'Block' = None):
-        self.name = name
-        self.status = "Success"
-        self.children = []
+    def __init__(self, name: str, parent: Optional[Block] = None):
```
i-keliukh, line 54: In order to use it with PyCharm, the name of the not yet defined class has to be expressed as a string literal: 'Block'

i-keliukh:
What is the proper way of running mypy on this?

universum/modules/vcs/vcs.py
```diff
@@ -20,35 +22,43 @@
 ]
 
 
-def create_vcs(class_type=None):
+def create_vcs(class_type: str = None) -> Type[ProjectDirectory]:
```
i-keliukh, line 25: Considering that the function is used to create base classes, we need to decide what errors we can catch with this annotation.

universum/modules/structure_handler.py
i-keliukh: The manual: https://www.python.org/dev/peps/pep-0484/#forward-references

universum/modules/structure_handler.py
o-andrieiev: could you please point to a line where it's defined?

universum/modules/structure_handler.py
o-andrieiev: ok, will fix

universum/modules/vcs/vcs.py
o-andrieiev: I'd say we can't do much with it right now due to https://github.com/python/mypy/issues/2477

universum/modules/structure_handler.py
i-keliukh: line 70

universum/modules/vcs/vcs.py
i-keliukh: I propose to first decide what is the goal of the type checking for this particular case and then maybe refactor vcs.py. For example, we can remove the dynamic base class by switching from the inheritance to composition.

PR #482, opened 2020-08-10
refactor: add type annotations to gravity module

o-andrieiev:
# Description

Without the fix, checking 'main' using mypy generates some errors. With the fix the only remaining errors are due to lack of static type annotations in the dependencies (third-party libraries).
To make static analysis of 'vcs' module meaningful, some minor refactoring was needed.
Also, added static type annotations highlighted some missing dependencies in 'git_vcs'.

!!! Please note that no tests have been run yet - this is preliminary pull request for some early feedback.

# Checklist:

- [x] My code follows the [PEP 8](https://www.python.org/dev/peps/pep-0008/)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/configs.py
+++ b/configs.py
@@ -32,7 +32,7 @@ configs = Variations([dict(name="Update Docker images", command=["make", "images
                                     "--rcfile=pylintrc", "--result-file=${CODE_REPORT_FILE}",
                                     "--files", "*.py", "universum/", "tests/"]),
                       dict(name="Run static type checker", code_report=True,
-                           command=["python3.7", "-m", "mypy", "universum/__main__.py", "--ignore-missing-imports"]),
+                           command=["python3.7", "-m", "mypy", "universum/", "--ignore-missing-imports"]),
 
                       dict(name="Run Jenkins plugin Java tests",
                            artifacts="universum_log_collapser/universum_log_collapser/target/surefire-reports/*.xml",
--- a/configs.py
+++ b/configs.py
@@ -31,6 +31,8 @@ configs = Variations([dict(name="Update Docker images", command=["make", "images
                            command=["python3.7", "-m", "universum.analyzers.pylint", "--python-version=3.7",
                                     "--rcfile=pylintrc", "--result-file=${CODE_REPORT_FILE}",
                                     "--files", "*.py", "universum/", "tests/"]),
+                      dict(name="Run static type checker", code_report=True,
+                           command=["python3.7", "-m", "mypy", "universum/main.py"]),
 
                       dict(name="Run Jenkins plugin Java tests",
                            artifacts="universum_log_collapser/universum_log_collapser/target/surefire-reports/*.xml",
--- a/configs.py
+++ b/configs.py
@@ -32,7 +32,7 @@ configs = Variations([dict(name="Update Docker images", command=["make", "images
                                     "--rcfile=pylintrc", "--result-file=${CODE_REPORT_FILE}",
                                     "--files", "*.py", "universum/", "tests/"]),
                       dict(name="Run static type checker", code_report=True,
-                           command=["python3.7", "-m", "mypy", "universum/main.py", "--ignore-missing-imports"]),
+                           command=["python3.7", "-m", "mypy", "universum/__main__.py", "--ignore-missing-imports"]),
 
                       dict(name="Run Jenkins plugin Java tests",
                            artifacts="universum_log_collapser/universum_log_collapser/target/surefire-reports/*.xml",
--- a/configs.py
+++ b/configs.py
@@ -32,7 +32,7 @@ configs = Variations([dict(name="Update Docker images", command=["make", "images
                                     "--rcfile=pylintrc", "--result-file=${CODE_REPORT_FILE}",
                                     "--files", "*.py", "universum/", "tests/"]),
                       dict(name="Run static type checker", code_report=True,
-                           command=["python3.7", "-m", "mypy", "universum/main.py"]),
+                           command=["python3.7", "-m", "mypy", "universum/main.py", "--ignore-missing-imports"]),
 
                       dict(name="Run Jenkins plugin Java tests",
                            artifacts="universum_log_collapser/universum_log_collapser/target/surefire-reports/*.xml",
--- a/universum/lib/gravity.py
+++ b/universum/lib/gravity.py
@@ -1,5 +1,5 @@
 from typing import cast, Any, Callable, ClassVar, Dict, Generic, List, Optional, Type, TypeVar, TYPE_CHECKING, Union
-
+from typing_extensions import Protocol
 __all__ = [
     "Module",
     "Dependency",
@@ -51,8 +51,9 @@ class Settings:
         setattr(obj.main_settings, self.cls.__name__, settings)
 
 
-class HasModulesMapping:
+class HasModulesMapping(Protocol):
     active_modules: Dict[Type['Module'], 'Module']
+
 
 class Module:
     settings: ClassVar['Settings']
--- a/universum/lib/gravity.py
+++ b/universum/lib/gravity.py
@@ -1,4 +1,4 @@
-from typing import cast, Any, Callable, ClassVar, Dict, Generic, List, NoReturn, Optional, Type, TypeVar, Union
+from typing import cast, Any, Callable, ClassVar, Dict, Generic, List, Optional, Type, TypeVar, Union
 
 __all__ = [
     "Module",
@@ -12,7 +12,6 @@ class ModuleSettings:
     def __init__(self, cls: Type['Module'], main_settings: 'ModuleSettings') -> None:
         object.__setattr__(self, "cls", cls)
         object.__setattr__(self, "main_settings", main_settings)
-        # TODO: can't clarify that active_modules[Type[X]] returns X (https://github.com/python/mypy/issues/4928)
         self.active_modules: Dict[Type[Module], Module]
 
     def __getattribute__(self, item: str) -> Union[str, List[str]]:
@@ -53,16 +52,18 @@ class Settings:
         setattr(obj.main_settings, self.cls.__name__, settings)
 
 
-class Module():
+class Module:
     settings: ClassVar['Settings']
     main_settings: 'ModuleSettings'
+
     def __new__(cls: Type['Module'], main_settings: 'ModuleSettings', *args, **kwargs) -> 'Module':
         instance: Module = super(Module, cls).__new__(cls)
         instance.main_settings = main_settings
         return instance
 
 
-ComponentType = TypeVar('ComponentType', bound='Module')
+ComponentType = TypeVar('ComponentType', bound=Module)
+
 
 def construct_component(cls: Type[ComponentType], main_settings: 'ModuleSettings', *args, **kwargs) -> ComponentType:
     if not getattr(main_settings, "active_modules", None):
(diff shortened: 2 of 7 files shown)


universum/lib/gravity.py
```diff
@@ -7,28 +9,30 @@
...
                 return getattr(settings, item)
             except AttributeError:
                 continue
 
         raise AttributeError("'" + cls.__name__ + "' object has no setting '" + item + "'")
 
-    def __setattr__(self, key, value):
-        cls = object.__getattribute__(self, "cls")
-        main_settings = object.__getattribute__(self, "main_settings")
+    def __setattr__(self, key: Any, value: Any) -> None:  # TODO: narrow down type annotations
```
o-andrieiev, line 30: I couldn't easily discern whether there is any type constraint to 'key' and 'value' arguments. Please let me know if there is

universum/lib/gravity.py
```diff
@@ -7,28 +9,30 @@
...
+        cls: Type['Module'] = object.__getattribute__(self, "cls")
+        main_settings: 'ModuleSettings' = object.__getattribute__(self, "main_settings")
         for entry in cls.__mro__:
             try:
-                settings = getattr(main_settings, entry.__name__)
+                settings: Any = getattr(main_settings, entry.__name__)  # TODO: narrow down type annotations
```
o-andrieiev, line 35: is this always a 'Settings' object? Not sure

universum/lib/gravity.py
```diff
@@ -7,28 +9,30 @@
 
 
 class ModuleSettings:
-    def __init__(self, cls, main_settings):
+    def __init__(self, cls: Type['Module'], main_settings: 'ModuleSettings') -> None:
         object.__setattr__(self, "cls", cls)
         object.__setattr__(self, "main_settings", main_settings)
+        # TODO: can't clarify that active_modules[Type[X]] returns X (https://github.com/python/mypy/issues/4928)
+        self.active_modules: Optional[Dict[Type[Module], Module]]
```
o-andrieiev, line 16: for some reason mypy lacks support of TypeVar in dicts. I can either leave it as is or refactor in a separate class with its own getter and setter. However that seems excessive, since 'construct_component' already enforces strict type adherence for return value, and 'active_modules' is not used outside of 'gravity' module.

universum/lib/gravity.py
k-dovgan: I think yes.

universum/lib/gravity.py
k-dovgan: I think, as these are the settings, `key` is always `str`, and `value` may be `str` or `List[str]`. But this might require further investigation.

o-andrieiev: thanks, I'll try these

universum/lib/gravity.py
o-andrieiev: thanks

configs.py
```diff
@@ -31,6 +31,8 @@ def pip_install(module_name):
                            command=["python3.7", "-m", "universum.analyzers.pylint", "--python-version=3.7",
                                     "--rcfile=pylintrc", "--result-file=${CODE_REPORT_FILE}",
                                     "--files", "*.py", "universum/", "tests/"]),
+                      dict(name="Run static type checker", code_report=True,
+                           command=["python3.7", "-m", "mypy", "universum/main.py", "--ignore-missing-imports"]),
```
o-andrieiev, line 35: 'glob2', 'sh', and 'urllib3' need stubs to remove the import errors
or we can just wait until their maintainers add type annotations themselves

universum/modules/structure_handler.py
```diff
@@ -10,16 +11,23 @@
...
-def needs_structure(klass):
-    klass.structure_factory = Dependency(StructureHandler)
-    original_init = klass.__init__
+class HasStructure():
+    structure_factory: Dependency['StructureHandler']
+    def __init__(self):
+        structure: 'StructureHandler'
+
+
+def needs_structure(cls: Type) -> Type['HasStructure']:
```
o-andrieiev, line 20: this loses information that resulting type is same class as incoming type, but preserves information about extra added fields. I'm not sure what would be the best way to preserve both.
The refactoring we discussed yesterday, where needs_structure becomes an interface/class resolves this issue.

universum/lib/gravity.py
i-keliukh: As an option, you can turn it into a pair of typed functions.

i-keliukh:
PyCharm produces a lot of warnings with current version of gravity.py.
Some of them are understandable and pretty easy to fix (number of CR/LF before function definition), but others are not.
Please check the code using PyCharm inspections.

universum/lib/gravity.py
o-andrieiev: yes, that's similar to the option 1 from the comment above. Still, I don't think it adds much value considering the 'construct_component'

i-keliukh: I agree.

universum/lib/gravity.py
```diff
@@ -7,28 +9,30 @@
 
 
 class ModuleSettings:
-    def __init__(self, cls, main_settings):
+    def __init__(self, cls: Type['Module'], main_settings: 'ModuleSettings') -> None:
```
i-keliukh, line 12: `main_settings` is `universum.lib.module_arguments.ModuleNamespace`, please see https://github.com/Samsung/Universum/blob/042ed269b1138273217ca2185a903498b3bcfd38/tests/test_gravity.py#L187-L191

universum/lib/gravity.py
```diff
@@ -39,69 +43,80 @@ def __setattr__(self, key, value):
 
 
 class Settings:
-    def __init__(self, cls):
-        self.cls = cls
+    def __init__(self, cls: Type['Module']) -> None:
+        self.cls: Type['Module'] = cls
 
-    def __get__(self, obj, objtype=None):
+    def __get__(self, obj: 'Module', objtype: Any = None) -> 'ModuleSettings':
```
i-keliukh, line 49: I am wondering, would it make more sense to make it explicit that ModuleSettings is just a proxy object for ModuleNamespace by using corresponding annotation?

o-andrieiev: hm, sounds interesting, but I don't think there is an easy way to do this. Let me think about it

universum/lib/gravity.py
o-andrieiev: will fix, my bad, all other main_settings seemed to point to ModuleSettings - which was pretty confusing. I think this annotation already makes it much clearer that ModuleSettings is a proxy to ModuleNamespace.

i-keliukh: All `main_settings` are instances of `ModuleNamespace`

i-keliukh:
Please also annotate test_gravity.py and include its verification into CI config.
Current version of the config doesn't verify tests.

universum/lib/gravity.py
```diff
@@ -39,69 +43,80 @@ def __setattr__(self, key, value):
...
-            return construct_component(self.klass, instance.main_settings, *args, **kwargs)
+    def __get__(self, instance: DependencyType, owner: Any) -> Callable[..., DependencyType]:
+        def constructor_function(*args, **kwargs) -> DependencyType:
+            return construct_component(self.cls, instance.main_settings, *args, **kwargs)
         return constructor_function
 
 
-def get_dependencies(klass, result=None, parent=None):
+def get_dependencies(cls: Type[Module], result: Optional[List[Type[Module]]] = None,
+                     parent: Optional[Dict[Type[Module], Type[Module]]] = None) -> List[Type[Module]]:
```
i-keliukh, line 95: You can safely remove `parent` argument instead of annotating it.
We should have removed it during the refactoring, but didn't do it by mistake.

universum/lib/module_arguments.py
```diff
@@ -1,11 +1,20 @@
 # pylint: disable = protected-access
 
+from typing import Dict, Type, TYPE_CHECKING
+
 import argparse
 import os
 import sys
 
+if TYPE_CHECKING:
+    from .gravity import Module  # for static type check
```
i-keliukh, line 10: I don't like cross-dependency between module_arguments and gravity.
Without type annotations, gravity can use any class instance as main_settings, but it is ok to explicitly state it is ModuleNamespace.
However, here we have backward dependency.
Could you, please, find a way to do it without cross dependencies?

universum/lib/gravity.py
```diff
@@ -1,3 +1,8 @@
+from typing import cast, Any, Callable, ClassVar, Generic, List, Optional, Type, TypeVar, TYPE_CHECKING, Union
+
+if TYPE_CHECKING:
```
i-keliukh, line 3: I think, it is ok to make the dependency unconditional.

universum/lib/module_arguments.py
o-andrieiev: I removed cross dependency by creating separate interface and inheriting it - similar to what we plan to do with 'needs_structure' and 'needs_output'

universum/lib/gravity.py
o-andrieiev: no longer seems to make sense, unless we want to closely couple ModuleNamespace and gravity

universum/lib/gravity.py
```diff
@@ -7,28 +9,28 @@
...
 
-    def __getattribute__(self, item):
-        cls = object.__getattribute__(self, "cls")
+    def __getattribute__(self, item: str) -> Union[str, List[str]]:
+        cls: Type['Module'] = object.__getattribute__(self, "cls")
         main_settings = object.__getattribute__(self, "main_settings")
         for entry in cls.__mro__:
             try:
-                settings = getattr(main_settings, entry.__name__)
+                settings: 'Settings' = getattr(main_settings, entry.__name__)
```
i-keliukh, line 21: This one is also incorrect.
All members of `main_settings`, except `active_modules`, `main_class` and `command_parser` are also `ModuleNamespace`

universum/lib/gravity.py
```diff
@@ -7,28 +9,28 @@
...
-    def __setattr__(self, key, value):
-        cls = object.__getattribute__(self, "cls")
-        main_settings = object.__getattribute__(self, "main_settings")
+    def __setattr__(self, key: str, value: Union[str, List[str]]) -> None:
+        cls: Type['Module'] = object.__getattribute__(self, "cls")
+        main_settings: 'HasModulesMapping' = object.__getattribute__(self, "main_settings")
         for entry in cls.__mro__:
             try:
-                settings = getattr(main_settings, entry.__name__)
+                settings: 'Settings' = getattr(main_settings, entry.__name__)
```
i-keliukh, line 33: Same issue as on line 21

universum/lib/gravity.py
```diff
@@ -7,28 +9,28 @@
...
 
-    def __getattribute__(self, item):
-        cls = object.__getattribute__(self, "cls")
+    def __getattribute__(self, item: str) -> Union[str, List[str]]:
+        cls: Type['Module'] = object.__getattribute__(self, "cls")
         main_settings = object.__getattribute__(self, "main_settings")
         for entry in cls.__mro__:
             try:
-                settings = getattr(main_settings, entry.__name__)
+                settings: 'HasModulesMapping' = getattr(main_settings, entry.__name__)
```
i-keliukh, line 21: I am sorry to say this, but while this member is `ModuleNamespace`, it doesn't in fact contain `active_modules`.
In general, I am not sure what benefits we can have from annotating it.
This is because it is designed to be dynamically populated from the command line arguments.
Probably, it can be fixed if we change `define_arguments` to static typing.

universum/modules/vcs/git_vcs.py
```diff
@@ -2,7 +2,7 @@
 import importlib
 import os
 
-from .base_vcs import BaseVcs, BaseDownloadVcs
+from .base_vcs import BaseVcs, BaseDownloadVcs, BasePollVcs, BaseSubmitVcs
```
o-andrieiev, line 5: this line generates PyCharm warning due to BaseVcs not being present in __all__

i-keliukh (approved):

PR #486, opened 2020-08-19
fix(gravity): incorrect type in Dependency

i-keliukh:
The Dependency class was annotated as using the same type for both
source and and target module of the dependency. Because of this
incorrect annotation, mypy produces the error at the place in the code,
where the source module calls the factory to construct the target
module.

Fixed by using two distinct type variables: one for source and one for
target.

Also, added type annotation in one method in submitter to properly
demonstrate the issue.

### DIFF
--- a/universum/lib/gravity.py
+++ b/universum/lib/gravity.py
@@ -96,7 +96,7 @@ class Dependency(Generic[DependencyType]):
     # The source and the target of the dependency are different modules,
     # so we need to use different type variables to annotate them.
     # instance parameter of the __get__ method is source module.
-    def __get__(self, instance: ComponentType, owner: Any) -> Callable[..., DependencyType]:
+    def __get__(self, instance: Module, owner: Any) -> Callable[..., DependencyType]:
         def constructor_function(*args, **kwargs) -> DependencyType:
             return construct_component(self.cls, instance.main_settings, *args, **kwargs)
         return constructor_function
--- a/universum/lib/gravity.py
+++ b/universum/lib/gravity.py
@@ -93,7 +93,10 @@ class Dependency(Generic[DependencyType]):
     def __init__(self, cls: Type[DependencyType]) -> None:
         self.cls = cls
 
-    def __get__(self, instance: DependencyType, owner: Any) -> Callable[..., DependencyType]:
+    # the source and the target of the dependency are different modules,
+    # so wee need to use different type variables to annotate them
+    # instance parameter of the __get__ method is source module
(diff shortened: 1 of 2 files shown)


universum/lib/gravity.py
```diff
@@ -93,7 +93,10 @@ class Dependency(Generic[DependencyType]):
     def __init__(self, cls: Type[DependencyType]) -> None:
         self.cls = cls
 
-    def __get__(self, instance: DependencyType, owner: Any) -> Callable[..., DependencyType]:
+    # the source and the target of the dependency are different modules,
+    # so wee need to use different type variables to annotate them
+    # instance parameter of the __get__ method is source module
+    def __get__(self, instance: ComponentType, owner: Any) -> Callable[..., DependencyType]:
```
o-andrieiev, line 99: does it add any value that we use 'ComponentType' here instead of just 'Module'?

universum/submit.py
```diff
@@ -30,8 +30,8 @@ def define_arguments(parser):
                             help="List of vcs or directories to be reconciled for commit. "
                                  "Relative paths starting at client root are supported")
 
-    def __init__(self, *args, **kwargs):
-        super(Submit, self).__init__(*args, **kwargs)
+    def __init__(self, *args, **kwargs) -> None:
```
o-andrieiev, line 33: does this belong to the same PR as the change above?

universum/lib/gravity.py
i-keliukh: I've checked it with the Module and it looks like it is also ok.

universum/submit.py
i-keliukh: Yes. As I mentioned in the description, this is added to illustrate the issue.
Without annotation on the client side, annotations in the gravity have no effect.

Also, I've tried to check if annotations give proper results.
It turned out that at least for vcs_factory() the mypy believes its type is ProjectDirectory.
Any attempt to use the result with real type (SubmitVcs) result it typing errors.
For instance, the following line produces error:
```python
change = self.vcs.driver.submit_new_change(self.settings.commit_message,
                                           path_list,
                                           review=self.settings.review,
                                           edit_only=self.settings.edit_only)
```

Another issue is that there is no possibility to annotate self.vcs with SubmitVcs, because it is a variable: https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases

o-andrieiev: Got it

o-andrieiev (approved):

Issue #261, opened 2019-02-20
Multiple errors in command-line parameters must be reported


i-keliukh:
Currently, if some command-line parameters are missing or incorrect, Universum reports the first error and stops. When user tries to launch the build, he is constantly faced with new errors appearing one after another.

Universum must collect all configuration errors as much as possible and report all of them. Of course, dependant errors must not be reported (e.g. no login and password for p4 lead to error in login and password in swarm).

The following tasks must be done in order to implement this task:
- [x] Analyze current checks
- [x] Add missing checks
- [x] Fix incorrect checks #271
- [x] Add tests for missing and incorrect checks
- [x] Change reporting of missing parameters to common mechanism
- [x] Implement mechanism for reporting error state for modules - classes derived from `Module`. Another option: register global error reporting function in 'Main'. Global function could be accessible via settings.
- [x] Use the above mentioned mechanism to report errors in input parameters and check error state in main function
- [x] Add tests for combinations of missing parameters with report of all of them
- [x] Add tests for single missing parameters but with different combination of other parameters (e.g. missing P4PORT with Jenkins and with TeamCity) - to make sure correct reporting is not affected by checks in other modules
- [x] Add tests for every parameter checked and every line changed by this feature
- [x] not needed as parameters are checked outside of Uninterruptible - ~~Refactor Uninterruptible and main classes to only use it from modules~~
- [x] not needed as parameters are checked outside of Uninterruptible - ~~Use error state mechanism in Uninterruptible~~

i-keliukh:
## Missing checks

The following checks are missing and must be implemented:
* P4CLIENT in submitter
* No P4_PATH and no P4_MAPPINGS 

## Checks to update

The following parameter checks must be rewritten to the error state:

- VCS: VCS type
- Git, Perforce: import errors
- Git: GIT_REPO
- Git: Code review (--report-to-review)
- Git submit: GITUSER, GITEMAIL
- Gerrit: not ssh url
- Gerrit: GIT_REFSPEC must be set
- Gerrit: GIT_CHECKOUT_ID must not be set
- Local VCS: SOURCE_DIR
- Local VCS: Code review (--report-to-review)
- P4: P4PORT, P4USER, P4PASSWD
- P4: submit with --create-review
- P4: P4_PATH and P4_MAPPINGS at the same time
- P4: P4CLIENT in PerforceMainVcs
- Swarm: REVIEW, SWARM_SERVER, SWARM_CHANGELIST
[...]

PR #487, opened 2020-08-20
feat(args): initial version of reporting multiple errors

i-keliukh:
This version only implements reporting multiple errors for the following parameters:
 - vcs type
 - config path
 - source path for local VCS

The idea is to commit this version and then gradually move the entire code base to this approach.

### DIFF
--- a/universum/modules/error_state.py
+++ b/universum/modules/error_state.py
@@ -1,4 +1,5 @@
 from typing import List
+import inspect
 
 from ..lib.gravity import Module, Dependency
 
@@ -18,7 +19,7 @@ class ErrorState(Module):
         self.global_error_state = self.global_error_state_factory()
 
     def error(self, message: str):
-        self.global_error_state.errors.append(message)
+        self.global_error_state.errors.append(inspect.cleandoc(message))
 
     def is_error_state(self):
         return len(self.global_error_state.errors) > 0
--- a/universum/modules/error_state.py
+++ b/universum/modules/error_state.py
@@ -9,6 +9,12 @@ class GlobalErrorState(Module):
         super().__init__(*args, **kwargs)  # type: ignore
         self.errors: List[str] = []
 
+    def get_errors(self):
+        return self.errors
+
+    def is_in_error_state(self):
+        return len(self.errors) > 0
+
 
 class HasErrorState(Module):
     global_error_state_factory = Dependency(GlobalErrorState)
@@ -21,7 +27,4 @@ class HasErrorState(Module):
         self.global_error_state.errors.append(inspect.cleandoc(message))
 
     def is_in_error_state(self):
-        return len(self.global_error_state.errors) > 0
-
-    def get_errors(self):
-        return self.global_error_state.errors
+        return self.global_error_state.is_in_error_state()
--- a/universum/modules/error_state.py
+++ b/universum/modules/error_state.py
@@ -10,18 +10,17 @@ class GlobalErrorState(Module):
         self.errors: List[str] = []
 
 
-class ErrorState(Module):
+class HasErrorState(Module):
     global_error_state_factory = Dependency(GlobalErrorState)
 
     def __init__(self, *args, **kwargs) -> None:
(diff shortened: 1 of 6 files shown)


i-keliukh:
Relates to #261

universum/modules/error_state.py
```diff
@@ -0,0 +1,28 @@
...
+from ..lib.gravity import Module, Dependency
+
+
+class GlobalErrorState(Module):
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)  # type: ignore
+        self.errors: List[str] = []
+
+
+class ErrorState(Module):
```
o-andrieiev, line 13: I don't like the name of this class. If we keep it as-is, it implies that Launches 'is a' ErrorState, which is wrong.
I propose 'HasErrorState' or 'ErrorStorage', or even something like 'ErrorCognizant'

universum/modules/error_state.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, *args, **kwargs) -> None:
+        # mypy doesn't support cooperative inheritance
+        super().__init__(*args, **kwargs)  # type: ignore
+        self.global_error_state = self.global_error_state_factory()
+
+    def error(self, message: str):
+        self.global_error_state.errors.append(inspect.cleandoc(message))
+
+    def is_error_state(self):
```
o-andrieiev, line 24: did you mean 'is_in_error_state'?

universum/modules/error_state.py
```diff
@@ -0,0 +1,28 @@
+from typing import List
+import inspect
+
+from ..lib.gravity import Module, Dependency
+
+
+class GlobalErrorState(Module):
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)  # type: ignore
```
o-andrieiev, line 9: codestyle question: should I use 'super()' in similar code, or 'super(GlobalErrorState, self)' ?

universum/modules/launcher.py
```diff
@@ -289,7 +289,7 @@ def _handle_postponed_out(self):
 
 @needs_output
 @needs_structure
-class Launcher(ProjectDirectory):
+class Launcher(ProjectDirectory, ErrorState):
```
o-andrieiev, line 292: codestyle/convention question: should ErrorState be inherited first or last?

universum/modules/error_state.py
i-keliukh: Use super().
super(GlobalErrorState, self) is only needed to support python 2.

universum/modules/error_state.py
i-keliukh: Well, I think it is fine either way.
However, if 'is_in_error_state' seems better to you, I will fix it.

universum/modules/launcher.py
i-keliukh: In python it only affects which members are shadowed.
If two bases are totally unrelated, the order doesn't matter.
I usually add a new base to the end of the list.
At least, this way I don't need to think about it too much.

universum/modules/error_state.py
i-keliukh: Ok.

universum/modules/error_state.py
o-andrieiev: I think 'is_in_error_state' is more clear

universum/modules/vcs/vcs.py
```diff
@@ -142,7 +144,8 @@ def __init__(self, *args, **kwargs):
         self.api_support = self.api_support_factory()
 
         if self.settings.report_to_review:
-            self.code_review = self.driver.code_review()
+            if not self.is_error_state():
```
o-andrieiev, line 147: wrong method name

i-keliukh: Looks like PyCharm missed it

o-andrieiev (approved):

PR #488, opened 2020-08-20
refactor: needs_output -> HasOutput

o-andrieiev:
# Description

Using needs_output to add .out attribute was a bit confusing,
and especially hard to align with static typing
Inheritance-based approach seems more natural (interface-like)
and should be more compatible with type annotations

# How Has This Been Tested?

Existing tests fully cover the change

# Checklist:

### DIFF
--- a/universum/modules/output/output.py
+++ b/universum/modules/output/output.py
@@ -4,21 +4,8 @@ from .terminal_based_output import TerminalBasedOutput
 from .teamcity_output import TeamcityOutput
 
 __all__ = [
-    "needs_output"
+    "HasOutput"
 ]
-
-
-def needs_output(klass):
-    klass.out_factory = Dependency(Output)
-    original_init = klass.__init__
-
-    def new_init(self, *args, **kwargs):
-        self.out = self.out_factory()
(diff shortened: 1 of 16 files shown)


o-andrieiev:
There're some redundant inheritances of HasOutput - I'll trim them down

universum/modules/output/output.py
```diff
@@ -72,3 +61,11 @@ def log_stderr(self, line):
 
     def log_shell_output(self, line):
         self.driver.log_shell_output(line)
+
+
+class HasOutput(Module):
+    out_factory: Dependency[Output] = Dependency(Output)
```
i-keliukh, line 67: Do you have to add annotation here?

o-andrieiev: not really, I just add them everywhere - better safe than sorry
I can remove this one, if you want

i-keliukh: I prefer to make sure that the inferred type is correct during development, but to remove repetition after that (DRY).

i-keliukh (approved):

PR #491, opened 2020-08-25
refactor: type annotations

o-andrieiev:
# Description

Add type annotations to 'main' and then resolve all missing dependencies
Some refactoring was needed

# How Has This Been Tested?

Existing tests are sufficient

# Checklist:

### DIFF
--- a/universum/main.py
+++ b/universum/main.py
@@ -81,8 +81,9 @@ class Main(HasOutput):
                     repo_diff = self.vcs.revert_repository()
                 except NotImplementedError:
                     self.out.log("Diff calculation for code report is skipped because current VCS doesn't support it")
-                self.launcher.launch_custom_configs(afterall_configs)
-                self.code_report_collector.repo_diff = repo_diff
+                else:
+                    self.launcher.launch_custom_configs(afterall_configs)
+                    self.code_report_collector.repo_diff = repo_diff
             self.code_report_collector.report_code_report_results()
         self.artifacts.collect_artifacts()
         self.reporter.report_build_result()
--- a/universum/main.py
+++ b/universum/main.py
@@ -79,10 +79,10 @@ class Main(HasOutput):
             if not self.settings.no_diff:
                 try:
                     repo_diff = self.vcs.revert_repository()
-                    self.launcher.launch_custom_configs(afterall_configs)
-                    self.code_report_collector.repo_diff = repo_diff
                 except NotImplementedError:
                     self.out.log("Diff calculation for code report is skipped because current VCS doesn't support it")
+                self.launcher.launch_custom_configs(afterall_configs)
+                self.code_report_collector.repo_diff = repo_diff
             self.code_report_collector.report_code_report_results()
         self.artifacts.collect_artifacts()
         self.reporter.report_build_result()
--- a/universum/main.py
+++ b/universum/main.py
@@ -1,22 +1,25 @@
+from typing import ClassVar
 from . import __title__
 from .lib.ci_exception import SilentAbortException
-from .lib.gravity import Module, Dependency
+from .lib.gravity import Dependency
+from .lib.module_arguments import ModuleArgumentParser
 from .modules import vcs, artifact_collector, reporter, launcher, code_report_collector
 from .modules.output import HasOutput
+
 
 __all__ = ["Main"]
 
 
 class Main(HasOutput):
-    description = __title__
-    vcs_factory = Dependency(vcs.MainVcs)
-    launcher_factory = Dependency(launcher.Launcher)
-    artifacts_factory = Dependency(artifact_collector.ArtifactCollector)
-    reporter_factory = Dependency(reporter.Reporter)
-    code_report_collector = Dependency(code_report_collector.CodeReportCollector)
+    description: ClassVar[str] = __title__
+    vcs_factory: ClassVar = Dependency(vcs.MainVcs)
+    launcher_factory: ClassVar = Dependency(launcher.Launcher)
+    artifacts_factory: ClassVar = Dependency(artifact_collector.ArtifactCollector)
+    reporter_factory: ClassVar = Dependency(reporter.Reporter)
+    code_report_collector_factory: ClassVar = Dependency(code_report_collector.CodeReportCollector)
 
     @staticmethod
-    def define_arguments(argument_parser):
+    def define_arguments(argument_parser: ModuleArgumentParser) -> None:
         argument_parser.add_hidden_argument("--no-finalize", action="store_true", dest="no_finalize",
                                             help="Skip 'Finalizing' step: "
                                                  "do not clear sources, do not revert workspace vcs, etc. "
@@ -41,14 +44,14 @@ class Main(HasOutput):
                                           "in this case full analysis report will be published")
 
     def __init__(self, *args, **kwargs):
-        super(Main, self).__init__(*args, **kwargs)
-        self.vcs = self.vcs_factory()
-        self.launcher = self.launcher_factory()
-        self.artifacts = self.artifacts_factory()
-        self.reporter = self.reporter_factory()
-        self.code_report_collector = self.code_report_collector()
+        super().__init__(*args, **kwargs)
+        self.vcs: vcs.MainVcs = self.vcs_factory()
+        self.launcher: launcher.Launcher = self.launcher_factory()
+        self.artifacts: artifact_collector.ArtifactCollector = self.artifacts_factory()
+        self.reporter: reporter.Reporter = self.reporter_factory()
+        self.code_report_collector: code_report_collector.CodeReportCollector = self.code_report_collector_factory()
 
-    def execute(self):
+    def execute(self) -> None:
         if self.settings.clean_build:
             self.vcs.clean_sources_silently()
             self.artifacts.clean_artifacts_silently()
@@ -74,17 +77,17 @@ class Main(HasOutput):
         self.launcher.launch_project()
         if afterall_configs:
             if not self.settings.no_diff:
-                if self.vcs.supports_copy_cl_files_and_revert():
+                try:
                     repo_diff = self.vcs.revert_repository()
                     self.launcher.launch_custom_configs(afterall_configs)
                     self.code_report_collector.repo_diff = repo_diff
-                else:
+                except NotImplementedError:
                     self.out.log("Diff calculation for code report is skipped because current VCS doesn't support it")
             self.code_report_collector.report_code_report_results()
         self.artifacts.collect_artifacts()
         self.reporter.report_build_result()
 
-    def finalize(self):
+    def finalize(self) -> None:
         if self.settings.no_finalize:
             self.out.log("Cleaning skipped because of '--no-finalize' option")
             return
--- a/universum/modules/artifact_collector.py
+++ b/universum/modules/artifact_collector.py
@@ -155,7 +155,7 @@ class ArtifactCollector(ProjectDirectory, HasOutput, HasStructure):
 
     @make_block("Preprocessing artifact lists")
     def set_and_clean_artifacts(self, project_configs, ignore_existing_artifacts=False):
-        artifact_list: List[Dict] = []
+        artifact_list = []
         report_artifact_list = []
         for configuration in project_configs.all():
             if "artifacts" in configuration:
--- a/universum/modules/artifact_collector.py
+++ b/universum/modules/artifact_collector.py
@@ -1,3 +1,5 @@
+from typing import List, Dict
+
 import codecs
 import distutils
 from distutils import dir_util, errors
@@ -6,7 +8,6 @@ import shutil
 import zipfile
 
 import glob2
-import six
 
 from ..lib.ci_exception import CriticalCiException, CiException
 from ..lib.gravity import Dependency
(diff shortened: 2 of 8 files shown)


universum/main.py
```diff
@@ -74,17 +77,17 @@ def execute(self):
         self.launcher.launch_project()
         if afterall_configs:
             if not self.settings.no_diff:
-                if self.vcs.supports_copy_cl_files_and_revert():
+                try:
```
o-andrieiev, line 80: This seems like a more pythonic approach

universum/modules/vcs/perforce_vcs.py
```diff
@@ -27,6 +30,11 @@
 P4Exception = None
 
 
+class P4stub(ModuleType):  # replace with proper stubs
+    P4Exception: Exception
+    P4: Callable
```
o-andrieiev, line 35: I know that this is a class, but then all attributes/methods of a class should also be stubbed.
Should I do that?

universum/modules/vcs/perforce_vcs.py
```diff
@@ -105,7 +113,7 @@ def disconnect(self):
                 return
             text = ""
             for line in w:
-                text += "\n" + warnings.formatwarning(line.message, line.category, line.filename, line.lineno)
+                text += "\n" + warnings.formatwarning(str(line.message), line.category, line.filename, line.lineno)
```
o-andrieiev, line 116: mypy was complaining that line.message is a Warning - not a str

universum/modules/vcs/perforce_vcs.py
```diff
@@ -499,34 +507,34 @@ def map_local_path_to_depot(self, report):
...
 
-        unshelved_filtered = [item for item in self.unshelved_files if item["action"] != "move/delete"]
+        unshelved_filtered: List[Dict[str, str]] =\
+            [item for item in self.unshelved_files if item["action"] != "move/delete"]
 
         for item in unshelved_filtered:
-            if item["action"] != "delete":
+            relative: Optional[str] = None
+            copied: Optional[str] = None
+            absolute: Optional[str]
```
o-andrieiev, line 520: this is to avoid re-declaration, which mypy doesn't like

universum/modules/vcs/perforce_vcs.py
```diff
@@ -499,34 +507,34 @@ def map_local_path_to_depot(self, report):
...
                 # for 'branch' diff we will assume it is a new file
                 # be careful, file for 'add' will be present in repo after revert
                 if item["action"] in ["add", "branch"]:
                     absolute = None
-                unshelved_path.append((relative, copied, absolute))
-
-            else:
-                absolute = os.path.join(self.client_root, item["clientFile"].replace("//" + item["client"] + "/", ""))
-                unshelved_path.append((None, None, absolute))
+            unshelved_path.append((relative, copied, absolute))
```
o-andrieiev, line 537: common code from both branches

universum/modules/vcs/perforce_vcs.py
```diff
@@ -499,34 +507,34 @@ def map_local_path_to_depot(self, report):
...
+        unshelved_filtered: List[Dict[str, str]] =\
+            [item for item in self.unshelved_files if item["action"] != "move/delete"]
 
         for item in unshelved_filtered:
-            if item["action"] != "delete":
+            relative: Optional[str] = None
+            copied: Optional[str] = None
+            absolute: Optional[str]
+            if item["action"] == "delete":
+                absolute = os.path.join(self.client_root, item["clientFile"].replace("//" + item["client"] + "/", ""))
```
o-andrieiev, line 522: shorter branch first

universum/main.py
i-keliukh: What if `launch_custom_configs` raises NotImplementedError exception?

universum/modules/vcs/perforce_vcs.py
i-keliukh: This is an external code.
We should not cover it, unless it gives us clear benefits.

universum/modules/vcs/perforce_vcs.py
i-keliukh: Can you check it?

universum/main.py
o-andrieiev: then previously that would stop the execution - > now it would just get looged as an unsupported action

universum/modules/vcs/perforce_vcs.py
o-andrieiev: I added these minimal stubs just to stop mypy from complaining (=

universum/modules/vcs/perforce_vcs.py
o-andrieiev: reveal_type(line.message)

> universum/modules/vcs/perforce_vcs.py:116: note: Revealed type is 'builtins.Warning'

universum/modules/vcs/vcs.py
```diff
@@ -172,10 +172,12 @@ def clean_sources_silently(self):
...
-        return self.driver.supports_copy_cl_files_and_revert
+    def copy_cl_files_and_revert(self) -> None:
+        raise NotImplementedError("Not applicable")
 
     @make_block("Revert repository")
-    def revert_repository(self):
-        diff = self.driver.copy_cl_files_and_revert()
-        return diff
+    def revert_repository(self) -> Optional[List[Tuple[Optional[str], Optional[str], Optional[str]]]]:
+        if isinstance(self.driver, base_vcs.BaseDownloadVcs):
```
i-keliukh, line 180: Is this pythonic?

universum/modules/vcs/perforce_vcs.py
i-keliukh: I don't understand what you are talking about, but it is fine if it is needed for mypy.
Maybe add a comment?

universum/modules/vcs/perforce_vcs.py
i-keliukh: How it was working before?
Is it some sort of automatic conversion to str inside `formatwarning`?

universum/modules/vcs/vcs.py
o-andrieiev: nope, but it calms mypy (=
I am still thinking about a good way to resolve this. I can do try catch and re-raise instead (like in 'main'), but then mypy would might want me to move 'copy_cl_files_and_revert' to common base class ('BaseDownloadVcs')

universum/main.py
i-keliukh: My point is that call to `launch_custom_configs` must be moved out of the try...except block, because with this change you log everything the same way.
Launcher should not be allowed to raise NotImplementedError, and the resulting error message must be different.

universum/modules/vcs/perforce_vcs.py
o-andrieiev: I think so

universum/modules/vcs/perforce_vcs.py
o-andrieiev: previously relative, copied, and absolute were declared inside branches
now they are declared before branching, updated inside branches, and used after branch
slighlty less code duplication

universum/modules/vcs/vcs.py
```diff
@@ -172,10 +172,12 @@ def clean_sources_silently(self):
         except OSError:
             pass
 
-    def supports_copy_cl_files_and_revert(self):
-        return self.driver.supports_copy_cl_files_and_revert
+    def copy_cl_files_and_revert(self) -> None:
```
i-keliukh, line 175: I don't think this function is needed here.

universum/modules/vcs/vcs.py
o-andrieiev: the problem is if this gets called when driver is not BaseDownloadVcs

universum/modules/vcs/vcs.py
o-andrieiev: removed

universum/modules/vcs/vcs.py
o-andrieiev: for now I moved 'copy_cl_files_and_revert' to BaseVcs to resolve this. It's always available, but in most cases causes NotImplementedError

universum/modules/vcs/vcs.py
```diff
@@ -172,12 +172,7 @@ def clean_sources_silently(self):
...
-    def copy_cl_files_and_revert(self) -> None:
-        raise NotImplementedError("Not applicable")
-
     @make_block("Revert repository")
     def revert_repository(self) -> Optional[List[Tuple[Optional[str], Optional[str], Optional[str]]]]:
-        if isinstance(self.driver, base_vcs.BaseDownloadVcs):
-            diff = self.driver.copy_cl_files_and_revert()
-            return diff
-        raise NotImplementedError("The driver is unable to perform the requested action")
+        diff = self.driver.copy_cl_files_and_revert()
```
i-keliukh, line 177: Do I understand it correctly: mypy doesn't know that driver has this function, so we had to move it to BaseVcs?

o-andrieiev: driver can be either DownloadVcs, SubmitVcs, or PollVcs - but only DownloadVcs has 'copy_cl_files_and_revert'.
This causes an issue in case this method is called for non-applicable driver.
By moving 'copy_cl_files_and_revert' to BaseVcs (common parent), we avoid the issue.

o-andrieiev: Of course - this is not the only way to resolve this. Checking whether the driver is fitting class also works - but that is less pythonic

o-andrieiev: or we can always silence the mypy warning

i-keliukh: I think the real issue is incorrect annotation of `driver` member if `MainVcs`.
`driver` cannot be SybmitVcs or PollVcs.

o-andrieiev: Got it. Fixed

universum/main.py
```diff
@@ -41,14 +44,14 @@ def define_arguments(argument_parser):
...
 
     def __init__(self, *args, **kwargs):
-        super(Main, self).__init__(*args, **kwargs)
-        self.vcs = self.vcs_factory()
-        self.launcher = self.launcher_factory()
-        self.artifacts = self.artifacts_factory()
-        self.reporter = self.reporter_factory()
-        self.code_report_collector = self.code_report_collector()
+        super().__init__(*args, **kwargs)
+        self.vcs: vcs.MainVcs = self.vcs_factory()
```
i-keliukh, line 48: Do we have to annotate it here?
My guess was that due to gravity being annotated mypy should be able to figure out types correctly.
Or this annotation works as double checks?

o-andrieiev: we don't have to annotate it here, but I prefer to annotate all self.blabla variables in __init__ as that is a natural place for people to look for attribute types

o-andrieiev: Plus double-check, as you mentioned. Even if gravity changes, we want these variables to keep their types.

i-keliukh: Ok, readability is enough reason to keep it.

universum/modules/artifact_collector.py
```diff
@@ -153,7 +155,7 @@ def preprocess_artifact_list(self, artifact_list, ignore_already_existing=False)
 
     @make_block("Preprocessing artifact lists")
     def set_and_clean_artifacts(self, project_configs, ignore_existing_artifacts=False):
-        artifact_list = []
+        artifact_list: List[Dict] = []
```
i-keliukh, line 158: In order to make this annotation work you need to annotate the method

universum/modules/code_report_collector.py
```diff
@@ -37,9 +39,9 @@ def prepare_environment(self, project_configs):
                 continue
 
             self.set_code_report_directory(self.settings.project_root)
-            temp_filename = "${CODE_REPORT_FILE}"
-            name = utils.calculate_file_absolute_path(self.report_path, item.get("name")) + ".json"
-            actual_filename = os.path.join(self.report_path, name)
+            temp_filename: str = "${CODE_REPORT_FILE}"
```
i-keliukh, line 42: Annotate the method to make this annotation work

universum/modules/code_report_collector.py
```diff
@@ -56,21 +58,22 @@ def prepare_environment(self, project_configs):
...
             with open(report_file, "r") as f:
-                text = f.read()
-                report = ""
+                text: str = f.read()
+                report: Optional[List[Dict[str, str]]] = None
                 if text:
                     report = json.loads(text)
 
-            json_file = self.artifacts.create_text_file("Static_analysis_report.json")
+            json_file: TextIO = self.artifacts.create_text_file("Static_analysis_report.json")
```
i-keliukh, line 70: Don't you have to import TextIO first?

universum/modules/artifact_collector.py
o-andrieiev: this looks a bit random. I wonder how it got there. I'll remove it for now, since it's not a dependency for other changes

universum/modules/code_report_collector.py
o-andrieiev: TextIO is imported above in line 2

i-keliukh: Sorry, somehow I didn't notice it.

i-keliukh (approved):

i-keliukh (approved):

PR #494, opened 2020-08-26
refactor: apply monkeytype annotations

o-andrieiev:
# Description

Run full pytest with monkeytype enabled
Apply monkeytype to all found modules
Manually fix all inconsistencies and erroneous imports
Add annotations to dependent classes

# How Has This Been Tested?

Covered by existing tests

# Checklist:


universum/api.py
```diff
@@ -1,50 +1,54 @@
+from types import TracebackType
+from typing import Optional
+
 import sys
 
 from .modules.api_support import ApiSupport
 from .lib.gravity import Module, Dependency
+from .lib.module_arguments import ModuleArgumentParser
```
o-andrieiev, line 8: ModuleArgumentParser is not annotated yet, that would be my next step

universum/api.py
```diff
@@ -1,50 +1,54 @@
...
             @staticmethod
-            def report_build_problem(problem):
+            def report_build_problem(problem: str) -> None:
                 pass
 
             @staticmethod
-            def log_exception(error):
-                ex_traceback = sys.exc_info()[2]
-                sys.stderr.write("Unexpected error.\n" + format_traceback(error, ex_traceback))
+            def log_exception(exc: Exception) -> None:
```
o-andrieiev, line 41: note: actual Output class uses 'str' here
maybe we should align the two

universum/configuration_support.py
```diff
@@ -213,10 +214,10 @@ def filter(self, checker, parent=None):
         if parent is None:
             parent = dict()
 
-        filtered_variations = []
+        filtered_variations: List = []
```
o-andrieiev, line 217: we can narrow it down to List[Dict]

universum/lib/gravity.py
```diff
@@ -11,12 +11,12 @@
 
 
 class ModuleSettings:
-    def __init__(self, cls: Type['Module'], main_settings: 'HasModulesMapping') -> None:
+    def __init__(self, cls, main_settings) -> None:
```
o-andrieiev, line 14: removed any annotations here due to lack of added value
mypy can't currently connect ModuleSettings to the underlying namespace, so type information is not preserved

universum/lib/utils.py
```diff
@@ -1,12 +1,17 @@
...
 import sys
 import traceback
 
 import requests
+from requests.models import Response
+from sh import RunningCommand
 
 from .ci_exception import CiException, CriticalCiException, SilentAbortException
 from .module_arguments import IncorrectParameterError
+from .gravity import ModuleSettings
```
o-andrieiev, line 14: ModuleSettings is not in __all__

universum/lib/utils.py
```diff
@@ -191,25 +205,25 @@ def excepted_function(func, *args, **kwargs):
                 self.return_code = max(self.return_code, 2)
         return excepted_function
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
```
o-andrieiev, line 208: unused params

universum/modules/automation_server/teamcity_server.py
```diff
@@ -87,18 +88,20 @@ def __init__(self, *args, **kwargs):
...
+    def artifact_path(self, local_artifacts_dir: str, item: str) -> str:
         return self.tc_artifact_link + item
 
-    def add_build_tag(self, tag):
+    def add_build_tag(self, tag: str) -> str:
         return requests.post("%s/httpAuth/app/rest/builds/id:%s/tags" %
                              (self.settings.server_url, self.settings.build_id),
                              auth=(self.settings.user_id, self.settings.passwd),
-                             data=tag, headers={'Content-Type': 'text/plain'}, verify=False)
+                             data=tag, headers={'Content-Type': 'text/plain'}, verify=False).text
```
o-andrieiev, line 107: do note ".text" here. It seems to me that str is expected, not Response

universum/modules/code_report_collector.py
```diff
@@ -14,6 +15,12 @@
 from .structure_handler import HasStructure
 
 
+class ProjectConfig(TypedDict, total=False):  # TODO: define a proper class for this
```
o-andrieiev, line 18: ProjectConfig seems like a complicated data structure, where value types for certain keys are important
consider refactoring as a full-fledged class

universum/modules/code_report_collector.py
```diff
@@ -23,37 +30,37 @@ def __init__(self, *args, **kwargs) -> None:
...
-            name = utils.calculate_file_absolute_path(self.report_path, item.get("name")) + ".json"
-            actual_filename = os.path.join(self.report_path, name)
+            temp_filename: str = "${CODE_REPORT_FILE}"
+            name: str = utils.calculate_file_absolute_path(self.report_path, item.get("name", "")) + ".json"
+            actual_filename: str = os.path.join(self.report_path, name)
 
             for key in item:
                 if key == "command":
-                    item[key] = [word.replace(temp_filename, actual_filename) for word in item[key]]
+                    item[key] = [word.replace(temp_filename, actual_filename) for word in item[key]]  # type: ignore
```
o-andrieiev, line 55: dynamic keys can't be used with TypedDict

universum/modules/code_report_collector.py
```diff
@@ -23,37 +30,37 @@ def __init__(self, *args, **kwargs) -> None:
...
                 else:
                     try:
-                        item[key] = item[key].replace(temp_filename, actual_filename)
+                        item[key] = item[key].replace(temp_filename, actual_filename)  # type: ignore
```
o-andrieiev, line 58: dynamic keys can't be used with TypedDict

universum/modules/code_report_collector.py
i-keliukh: project_configs is just Variations merged into a flat list. Hence, ProjectConfig is just one configuration.
Its keys are described here: https://universum.readthedocs.io/en/latest/configuring.html#common-variations-keys

o-andrieiev: how about we define all of the possible keys in ProjectConfig TypedDict?
Would that be useful?

i-keliukh: Yes, for sure.
Can we somehow connect it to documentation?
For example, is there possibility to annotate and document it within the code?

i-keliukh: The only thing, this is not the place. This module is just a sidekick of a Launcher.

o-andrieiev: > Can we somehow connect it to documentation?
> For example, is there possibility to annotate and document it within the code?
I need to investigate that. This ProjectConfig(TypedDict) mypy class doesn't exist in runtime (it's just dict), but we might still put docstrings there.

o-andrieiev: > The only thing, this is not the place. This module is just a sidekick of the Launcher.

I was considering putting the new class to Variations, but that one seems more generic. Launcher is a fine location too, I guess.

o-andrieiev: Should I do this as part of current PR?

i-keliukh: > I was considering putting the new class to Variations, but that one seems more generic. Launcher is a fine location too, I guess.

Variations is more user friendly. Launcher is the one that mostly uses those keys.
I think we should be user friendly.

I agree that Variations is generic, but, frankly, no one will ever use it for anything except creating configs for Launcher.
Also, I don't think we will ever see anything, but Variations, passed as configs to Launcher.

i-keliukh: > Should I do this as part of current PR?

I think it deserves separate commit.

o-andrieiev: I think so too, especially if I am going to narrow down Variations (which seems like a good idea to me now, but we'll see)

o-andrieiev: > I need to investigate that. This ProjectConfig(TypedDict) mypy class doesn't exist in runtime (it's just dict), but we might still put docstrings there.

Tested and confirmed that sphinx works even for helper classes like this.

universum/api.py
```diff
@@ -1,50 +1,54 @@
...
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         try:
-            self.api_support = self.api_support_factory(api_mode=True)
-        except EnvironmentError as error:
-            sys.stderr.write(str(error) + '\n')
+            self.api_support: ApiSupport = self.api_support_factory(api_mode=True)
+        except EnvironmentError as e:
```
k-dovgan, line 27: Why rename into 'e'? Also, doesn't pylint have warnings on too short var names?

universum/lib/utils.py
```diff
@@ -26,14 +31,17 @@
     "make_request"
 ]
 
+ReturnT = TypeVar('ReturnT')
```
k-dovgan, line 34: Why not ReturnType and DecoratorType?

universum/lib/utils.py
k-dovgan: context manager convention

universum/modules/api_support.py
```diff
@@ -24,25 +25,25 @@ def __init__(self, api_mode=False, **kwargs):
                     please don't. If you got this message by running it with Universum: something must
                     have gone wrong, may be a bug in Universum itself. Feel free to contact the developers."""))
 
-            with open(os.getenv("UNIVERSUM_DATA_FILE"), "rb+") as data_file:
+            with open(os.getenv("UNIVERSUM_DATA_FILE", ""), "rb+") as data_file:
```
k-dovgan, line 28: This way it won't raise an exception if the env var somehow is absent. And it should.

universum/modules/automation_server/teamcity_server.py
k-dovgan: Where did the '.text' come from? It returned full Response before this change

k-dovgan (commented):
Can you, maybe, move all fixes not related to typing, to a separate commit/PR?

universum/api.py
o-andrieiev: e is canonical for errors, like i for indices.  I'm not sure exactly why I changed it here - I can revert this change

universum/modules/api_support.py
o-andrieiev: then this code should be rewritten. Relying on 'open' failing when invalid input is given is not a good practice, I think.

universum/api.py
k-dovgan: Sure, it definitely should expect string.

universum/modules/api_support.py
k-dovgan: Sure, but it's a change to be done in a separate PR, isn't it?

universum/modules/automation_server/teamcity_server.py
o-andrieiev: I wasn't sure about this one - which is exactly why I added this comment.

universum/lib/utils.py
o-andrieiev: no special reason, just the convention I've seen in mypy docs.
Would you prefer Type?

universum/lib/utils.py
```diff
@@ -68,7 +76,15 @@ def detect_environment():
     return "terminal"
 
 
-def create_driver(local_factory, teamcity_factory, jenkins_factory, env_type=""):
+T1 = TypeVar('T1')
```
o-andrieiev, line 79: please feel free to suggest better naming here

universum/modules/api_support.py
o-andrieiev: I think minor changes (like the one needed here) can be done in the same PR.

o-andrieiev: actually, I don't think any change is needed here, since open can't process "" as a filename anyway

universum/api.py
o-andrieiev: I'll change it in a separate PR

universum/configuration_support.py
```diff
@@ -128,7 +129,7 @@ def __add__(self, other):
         """
         return Variations(list.__add__(list(self), other))
 
-    def __mul__(self, other):
+    def __mul__(self, other: Union[int, List]) -> 'Variations':
```
i-keliukh, line 132: Why Union[int, List], not Union[int, 'Variations']?

i-keliukh: The doc string has to include `int` as valid type for `other`

universum/configuration_support.py
```diff
@@ -155,7 +156,7 @@ def __mul__(self, other):
         result = Variations(result_list)
         return result
 
-    def all(self):
+    def all(self) -> Generator[Dict, None, None]:
```
i-keliukh, line 159: Wouldn't it be better to use Iterable[Dict]?

universum/lib/utils.py
i-keliukh: T1, T2 and T3 are any Modules.
However, they are definitely for local environment, teamcity environment and jenkins environment.

universum/lib/utils.py
```diff
@@ -79,18 +95,18 @@ def create_driver(local_factory, teamcity_factory, jenkins_factory, env_type="")
...
-    tb_lines = traceback.format_exception(exc.__class__, exc, trace)
-    tb_text = ''.join(tb_lines)
+def format_traceback(exc: Exception, trace: Optional[TracebackType]) -> str:
+    tb_lines: List[str] = traceback.format_exception(exc.__class__, exc, trace)
+    tb_text: str = ''.join(tb_lines)
     return tb_text
 
 
-def check_required_option(settings, setting_name, error_message):
+def check_required_option(settings: ModuleSettings, setting_name: str, error_message: str) -> None:
```
i-keliukh, line 104: It should be ModuleNamespace

universum/lib/utils.py
o-andrieiev: Jenkins factory can either be for JenkinsServerForHostingBuild or for JenkinsServerForTrigger. I am fine with hardcoding acceptable factory types here, but it seems excessive to me. E.g. the way it's written right now we can add other factories for local/teamcity/jenkins without changing this util function.

universum/lib/utils.py
```diff
@@ -79,18 +95,18 @@ def create_driver(local_factory, teamcity_factory, jenkins_factory, env_type="")
...
     if not getattr(settings, setting_name, None):
         raise IncorrectParameterError(inspect.cleandoc(error_message))
 
 
-def read_and_check_multiline_option(settings, setting_name, error_message):
+def read_and_check_multiline_option(settings: ModuleSettings, setting_name: str, error_message: str) -> str:
```
i-keliukh, line 109: Same issue

universum/lib/utils.py
```diff
@@ -131,7 +147,7 @@ def function_to_run(*args, **kwargs):
     return decorated_function
 
 
-def trim_and_convert_to_unicode(line):
+def trim_and_convert_to_unicode(line: Union[bytes, str, RunningCommand]) -> str:
```
i-keliukh, line 150: RunningCommand? That is surprising.
Could you, please, convert it to string on the client side?

universum/lib/utils.py
o-andrieiev: changing ModuleSettings to ModuleNamespace here results in mypy error. Investigating

universum/configuration_support.py
```diff
@@ -116,30 +117,30 @@ class Variations(list):
 
     """
 
-    def __add__(self, other):
+    def __add__(self, other: List) -> 'Variations':
```
o-andrieiev, line 120: typing 'other' as 'Variations' here generates mypy error:
universum/configuration_support.py:120: error: Argument 1 of "__add__" is incompatible with supertype "list"; supertype defines the argument type as "List[Any]"
universum/configuration_support.py:120: note: This violates the Liskov substitution principle
universum/configuration_support.py:120: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

universum/lib/utils.py
o-andrieiev: changed to HasModulesMapping Protocol

universum/lib/utils.py
o-andrieiev: changed to HasModulesMapping Protocol

universum/configuration_support.py
o-andrieiev: I used List to align with __add__ parameters
'Variations' can always be used instead of List

o-andrieiev: Changed to 'Variations' now

universum/lib/utils.py
o-andrieiev: sure

o-andrieiev: is 'bytes' expected?

universum/modules/automation_server/automation_server.py
```diff
@@ -26,20 +29,21 @@ class AutomationServerForHostingBuild(AutomationServer):
...
     jenkins_driver_factory = Dependency(JenkinsServerForHostingBuild)
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
-        self.driver = utils.create_driver(teamcity_factory=self.teamcity_driver_factory,
-                                          jenkins_factory=self.jenkins_driver_factory,
-                                          local_factory=self.local_driver_factory,
-                                          env_type=self.settings.type)
+        self.driver: Union[LocalServer, TeamcityServer, JenkinsServerForHostingBuild] = \
```
i-keliukh, line 34: Can we use base class here instead?

universum/lib/utils.py
o-andrieiev: I think monkeytype just got confused by cmd.process.ran in 'launcher'.
I added some annotations there to make things clear.

universum/modules/automation_server/automation_server.py
o-andrieiev: sure, good idea

universum/lib/utils.py
i-keliukh: I was only talking about TypeVar names.
They are factories for known environment, so you can use more specific names instead of T1, T2 and T3.
Also, you can bind them to Module.

universum/lib/utils.py
i-keliukh: But they didn't help to resolve the issue?

universum/configuration_support.py
i-keliukh: I don't see the change.

universum/lib/utils.py
i-keliukh: ReturnT is ok to me.

universum/configuration_support.py
o-andrieiev: yep. We override the method, but narrow down the expected argument, which means that Variations can't seamlessly replace instances of List, which violates Liskov subs. principle

o-andrieiev: that's why initially I used List for __mul__ too, since I assumes it should work for any List

universum/configuration_support.py
o-andrieiev: Please double-check whether you're reviewing the latest revision

i-keliukh: You are right. I didn't see the last version.

universum/lib/utils.py
o-andrieiev: mypy doesn't see any issues now, so I guess it's resolved

i-keliukh: Please ignore the last comment. I didn't see the change.

i-keliukh (approved):

PR #495, opened 2020-08-26
feat(args): printing multiple errors in p4, submit

i-keliukh:
Also removed tests with deleting fields from settings objects. If the
fields is absent, the error state mechanism is able to detect it and
remember the error. However, the code in the module still can access the
field. Such accesses result in exceptions raised from gravity.

This is not an issue for normal operation, because when user launches
the script, all settings fields are present.

### DIFF
--- a/universum/modules/error_state.py
+++ b/universum/modules/error_state.py
@@ -23,7 +23,7 @@ class HasErrorState(Module):
         super().__init__(*args, **kwargs)  # type: ignore
         self.global_error_state: GlobalErrorState = self.global_error_state_factory()
 
-    def error(self, message: str):
+    def error(self, message: str) -> None:
         self.global_error_state.errors.append(inspect.cleandoc(message))
 
     def is_in_error_state(self) -> bool:
--- a/universum/modules/error_state.py
+++ b/universum/modules/error_state.py
@@ -29,6 +29,6 @@ class HasErrorState(Module):
     def is_in_error_state(self) -> bool:
         return self.global_error_state.is_in_error_state()
 
-    def check_required_option(self, setting_name, message) -> None:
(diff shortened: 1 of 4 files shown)


i-keliukh:
Relates to #261

universum/modules/error_state.py
```diff
@@ -21,10 +21,14 @@ class HasErrorState(Module):
...
+        self.global_error_state: GlobalErrorState = self.global_error_state_factory()
 
     def error(self, message: str):
         self.global_error_state.errors.append(inspect.cleandoc(message))
 
-    def is_in_error_state(self):
+    def is_in_error_state(self) -> bool:
         return self.global_error_state.is_in_error_state()
+
+    def check_required_option(self, setting_name, message) -> None:
```
o-andrieiev, line 32: why did you omit types in setting_name and message?

i-keliukh: Is there some mode for mypy to warn about this?

o-andrieiev: --disallow-incomplete-defs

universum/modules/error_state.py
```diff
@@ -21,10 +21,14 @@ class HasErrorState(Module):
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)  # type: ignore
-        self.global_error_state = self.global_error_state_factory()
+        self.global_error_state: GlobalErrorState = self.global_error_state_factory()
 
     def error(self, message: str):
```
o-andrieiev, line 26: -> None missing

k-dovgan (approved):

k-dovgan (approved):

PR #496, opened 2020-08-26
fix(poll): exception if polling with local vcs

### LINKED ISSUE #493: poll generates error with -vt=none
## Steps to reproduce

Run the following command:

```bash
python3.7 -m universum poll -vt=none
```

## Expected result

Application finishes execution with error.

## Actual result

Application produces the exception backtrace:

==> Universum 0.18.7 started execution
1. Enumerating changes
 └ [Success]

2. Finalizing
 └ [Success]

```
Traceback (most recent call last):
  File "/home/<user>/Projects/Universum/universum/lib/utils.py", line 182, in excepted_function
    func(*args, **kwargs)
  File "/home/<user>/Projects/Universum/universum/lib/utils.py", line 205, in function_in_block
    return self.structure.run_in_block(func, block_name, pass_errors, self, *args, **kwargs)
  File "/home/<user>/Projects/Universum/universum/modules/structure_handler.py", line 116, in run_in_block
    result = operation(*args, **kwargs)
  File "/home/<user>/Projects/Universum/universum/poll.py", line 65, in execute
    self.latest_cls = self.vcs.driver.get_changes(self.stored_cls, self.settings.max_number)
  File "/home/<user>/Projects/Universum/universum/modules/vcs/base_vcs.py", line 95, in get_changes
    raise NotImplementedError
NotImplementedError
==> Universum 0.18.7 finished execution
```

i-keliukh:
Fixes #493.

### DIFF
--- a/universum/poll.py
+++ b/universum/poll.py
@@ -1,6 +1,6 @@
 import json
 
-from .lib.gravity import Module, Dependency
+from .lib.gravity import Dependency
 from .lib.utils import make_block
 from .modules import automation_server, vcs
 from .modules.output import HasOutput
@@ -65,6 +65,8 @@ class Poll(HasOutput, HasStructure):
             self.latest_cls = self.vcs.driver.get_changes(self.stored_cls, self.settings.max_number)
             for depot in self.latest_cls.keys():
                 self.structure.run_in_block(self.process_single_mapping, "Processing depot " + depot, True, depot)
+        except NotImplementedError:
+            self.out.log("Polling is skipped because current VCS doesn't support it")
         finally:
             with open(self.settings.db_file, "w") as db_file:
                 json.dump(self.stored_cls, db_file, indent=4, sort_keys=True)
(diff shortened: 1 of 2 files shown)


universum/poll.py
```diff
@@ -65,6 +65,8 @@ def execute(self):
             self.latest_cls = self.vcs.driver.get_changes(self.stored_cls, self.settings.max_number)
             for depot in self.latest_cls.keys():
                 self.structure.run_in_block(self.process_single_mapping, "Processing depot " + depot, True, depot)
+        except NotImplementedError:
+            self.out.log("Polling is skipped because current VCS doesn't support it")
```
o-andrieiev, line 69: isn't this Warning worthy? should we use log_stderr here?

o-andrieiev (approved):
I have a small suggestion, but feel free to ignore it. Looks good to me

universum/poll.py
i-keliukh: It is similar to the message with diff calculation for local vcs. IMO it is better to show the message than to silently do nothing.

o-andrieiev: I see

k-dovgan (approved):

PR #499, opened 2020-08-31
refactor: ProjectConfig and its usage

o-andrieiev:
Introduce new type for ProjectConfig
This does not introduce any runtime overhead, and is mostly needed for static typing
Text description from 'configuring' should be moved here eventually

### DIFF
--- a/doc/configuring.rst
+++ b/doc/configuring.rst
@@ -224,7 +224,7 @@ artifacts
     Path to the file or directory to be copied to the working directory as an execution result.
     Can contain shell-style pattern matching (e.g. `"out/*.html"`), including recursive wildcards
     (e.g. `"out/**/index.html"`).
-    If not stated otherwise (see ``--no-archive`` `command-line parameter <args.html#Artifact\ collection>`__
+    If not stated otherwise (see ``--no-archive`` `command-line parameter <args.html#Artifact\\ collection>`__
     for details), artifact directories are copied as archives.
     If :ref:`'artifact_prebuild_clean' key <clean_artifacts>` is either absent or set to `False` and
     stated artifacts are present in downloaded sources, it is considered a failure and configuration
@@ -260,9 +260,9 @@ artifact_prebuild_clean
 
 directory
     Path to a current working directory for launched process.
-    Please see the `Execution directory`_ section for details. No `directory` is equal to
-    empty string passed as `directory` and means the `command` will be launched
-    from project root directory.
+    Please see the `execution directory section <configuring.html#execution-directory>`__ for details. No `directory`
+    is equal to empty string passed as `directory` value and means the `command` will be launched from project
+    root directory.
 
 .. _critical_step:
 
@@ -311,11 +311,12 @@ pass_tag, fail_tag
     TeamCity as tags. It is not recommended to separate words in the tag with spaces, since you cannot create
     a tag with spaces in TeamCity's web-interface. Every tag is added (if matching condition) after executing
     build step it is set in, not in the end of all run.
-    ``pass_tag`` and ``fail_tag`` can also be used when `multiplying build configurations`_, like this:
+    ``pass_tag`` and ``fail_tag`` can also be used when
+    `multiplying build configuration <configuring.html#multiplying-build-configurations>`__, like this:
 
     .. testsetup::
 
-        #!/usr/bin/env {python}
+        #!/usr/bin/env python3.7
 
         from universum.configuration_support import Variations
 
--- a/doc/configuring.rst
+++ b/doc/configuring.rst
@@ -224,7 +224,7 @@ artifacts
     Path to the file or directory to be copied to the working directory as an execution result.
     Can contain shell-style pattern matching (e.g. `"out/*.html"`), including recursive wildcards
     (e.g. `"out/**/index.html"`).
-    If not stated otherwise (see ``--no-archive`` `command-line parameter <args.html#Artifact\\ collection>`__
+    If not stated otherwise (see ``--no-archive`` `command-line parameter <args.html#Artifact\ collection>`__
     for details), artifact directories are copied as archives.
     If :ref:`'artifact_prebuild_clean' key <clean_artifacts>` is either absent or set to `False` and
     stated artifacts are present in downloaded sources, it is considered a failure and configuration
--- a/doc/configuring.rst
+++ b/doc/configuring.rst
@@ -316,7 +316,7 @@ pass_tag, fail_tag
 
     .. testsetup::
 
-        #!/usr/bin/env python3.7
+        #!/usr/bin/env {python}
 
         from universum.configuration_support import Variations
 
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -4,11 +4,11 @@ import getpass
 import os
 from pwd import getpwnam
 
+from typing import Generator
 import docker
 import py
 import pytest
 from requests.exceptions import ReadTimeout
-from typing import Generator
 from . import utils
 from .utils import python
 
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -9,6 +9,7 @@ import docker
 import py
 import pytest
 from requests.exceptions import ReadTimeout
+
 from . import utils
 from .utils import python
 
--- a/tests/deployment_utils.py
+++ b/tests/deployment_utils.py
@@ -8,7 +8,7 @@ import docker
 import py
 import pytest
 from requests.exceptions import ReadTimeout
-
+from typing import Generator
 from . import utils
 from .utils import python
 
@@ -129,7 +129,7 @@ class ExecutionEnvironment:
 
 
 @pytest.fixture()
-def execution_environment(request) -> ExecutionEnvironment:
+def execution_environment(request) -> Generator[ExecutionEnvironment, None, None]:
     runner = None
     try:
         runner = ExecutionEnvironment(request, os.getcwd())
@@ -262,7 +262,7 @@ class UniversumRunner:
 
 
 @pytest.fixture()
-def runner_without_environment(perforce_workspace, git_client, local_sources) -> UniversumRunner:
+def runner_without_environment(perforce_workspace, git_client, local_sources) -> Generator[UniversumRunner, None, None]:
     runner = UniversumRunner(perforce_workspace, git_client, local_sources, nonci=False)
     yield runner
     runner.clean_artifacts()
--- a/universum/configuration_support.py
+++ b/universum/configuration_support.py
@@ -55,16 +55,15 @@ class ProjectConfiguration:
         for key, value in kwargs.items():
             self.extras[key] = value
 
-    def replace_string(self, from_string: str, to_string: str) -> None:
-        self.command = [word.replace(from_string, to_string) for word in self.command]
-        self.artifacts = self.artifacts.replace(from_string, to_string)
-        self.report_artifacts = self.report_artifacts.replace(from_string, to_string)
-        self.directory = self.directory.replace(from_string, to_string)
-        for k, v in self.extras.items():
-            if isinstance(v, str):
-                self.extras[k] = v.replace(from_string, to_string)
-
     def __repr__(self) -> str:
+        """
+        This function simulates dict-like legacy representation for output
+        :return: dict-like string
+
+        >>> cfg = ProjectConfiguration(name='foo', command=['bar'], my_var='baz')
+        >>> repr(cfg)
+        "{'name': 'foo', 'command': 'bar', 'my_var': 'baz'}"
+        """
         res = {k: v for k, v in self.__dict__.items() if v and k != 'extras'}
         res.update(self.extras)
         if len(self.command) == 1:  # command should be printed as one string, instead of list
@@ -72,17 +71,63 @@ class ProjectConfiguration:
         return str(res)
 
     def __eq__(self, other: Any) -> bool:
+        """
+        This functions simulates dict-like legacy comparison
+
+        :param other: dict to compare values, or object to check exact equality
+        :return: 'True' if 'other' matches
+
+        >>> cfg1 = ProjectConfiguration(name='foo', my_var='bar')
+        >>> cfg2 = ProjectConfiguration(name='foo', my_var='bar')
+        >>> cfg1 == cfg1
+        True
+        >>> cfg1 == cfg2
+        False
+        >>> cfg1 == {'name': 'foo', 'my_var': 'bar'}
+        True
+        >>> cfg1 == {'name': 'foo', 'my_var': 'bar', 'critical': False}
+        True
+        >>> cfg1 == {'name': 'foo', 'my_var': 'bar', 'test': None}
+        True
+        >>> cfg1 == {'name': 'foo', 'my_var': 'bar', 'test': ''}
+        False
+        """
         if isinstance(other, dict):
             for key, val in other.items():
-                if getattr(self, key, None) != val:
+                if self[key] != val:
                     return False
             return True
         return super().__eq__(other)
 
     def __getitem__(self, item: str) -> Optional[str]:
-        return self.extras.get(item, None)
+        """
+        This functions simulates dict-like legacy access
+
+        :param item: client-defined item
+        :return: client-defined value
+
+        >>> cfg = ProjectConfiguration(name='foo', my_var='bar')
+        >>> cfg['name']
+        'foo'
+        >>> cfg['my_var']
+        'bar'
+        >>> cfg['test']
+        """
+        return self.extras.get(item, self.__dict__.get(item, None))
 
     def __add__(self, other: 'ProjectConfiguration') -> 'ProjectConfiguration':
+        """
+        This functions defines operator ``+`` for :class:`.ProjectConfiguration` class objects by
+        concatenating strings and contents of dictionaries
+
+        :param other: `ProjectConfiguration` object
+        :return: new `ProjectConfiguration` object, including all attributes from both `self` and `other` objects
+
+        >>> cfg1 = ProjectConfiguration(name='foo', command=['foo'], critical=True, my_var1='foo')
+        >>> cfg2 = ProjectConfiguration(name='bar', command=['bar'], background=True, my_var1='bar', my_var2='baz')
+        >>> cfg1 + cfg2
+        {'name': 'foobar', 'command': ['foo', 'bar'], 'background': True, 'my_var1': 'foobar', 'my_var2': 'baz'}
+        """
         return ProjectConfiguration(
             name=self.name + other.name,
             command=self.command + other.command,
@@ -100,7 +145,46 @@ class ProjectConfiguration:
             extras=combine(self.extras, other.extras)
         )
 
+    def replace_string(self, from_string: str, to_string: str) -> None:
+        """
+        Replace instances of a string, used for pseudo-variables
+
+        :param from_string: string to replace, e.g. ${CODE_REPORT_FILE}
+        :param to_string: value to put in place of 'from_string'
+
+        >>> cfg = ProjectConfiguration(name='foo test', command=['foo', 'baz', 'foobar'], \
+        myvar1='foo', myvar2='bar', myvar3=1)
+        >>> cfg.replace_string('foo', 'bar')
+        >>> cfg
+        {'name': 'foo test', 'command': ['bar', 'baz', 'barbar'], 'myvar1': 'bar', 'myvar2': 'bar', 'myvar3': 1}
+
+        >>> cfg = ProjectConfiguration(artifacts='foo', report_artifacts='foo', directory='foo')
+        >>> cfg.replace_string('foo', 'bar')
+        >>> cfg
+        {'directory': 'bar', 'artifacts': 'bar', 'report_artifacts': 'bar'}
+        """
+        self.command = [word.replace(from_string, to_string) for word in self.command]
+        self.artifacts = self.artifacts.replace(from_string, to_string)
+        self.report_artifacts = self.report_artifacts.replace(from_string, to_string)
+        self.directory = self.directory.replace(from_string, to_string)
+        for k, v in self.extras.items():
+            if isinstance(v, str):
+                self.extras[k] = v.replace(from_string, to_string)
+
     def stringify_command(self) -> bool:
+        """
+        Concatenates components of a command into one element
+
+        :return: 'True', if any of the command components have space inside
+
+        >>> cfg = ProjectConfiguration(name='stringify test', command=['foo', 'bar', '--baz'])
+        >>> cfg.stringify_command()
+        False
+        >>> cfg
+        {'name': 'stringify test', 'command': 'foo bar --baz'}
+        >>> cfg.stringify_command()
+        True
+        """
         result = False
         command_line = ""
         for argument in self.command:
--- a/universum/configuration_support.py
+++ b/universum/configuration_support.py
@@ -1,4 +1,5 @@
 from typing import Any, Callable, Dict, Iterable, List, Optional, TypeVar, Union
+from warnings import warn
 import copy
 import os
 
@@ -33,8 +34,19 @@ class ProjectConfiguration:
                  pass_tag: str = '',
                  fail_tag: str = '',
                  if_env_set: str = '',
-                 extras: Optional[Dict[str, str]] = None,
-                 **kwargs) -> None:  # explicit constructor for client's type safety
+                 **kwargs) -> None:
+        """
+        All params supplied via named attributed shall be  type-checked for safety. Clients may provide custom
+        parameters in kwargs - these parameters will be stored internally in '_extras' dict and can be retrieved by
+        indexing.
+        >>> cfg = ProjectConfiguration(name='foo', command=['1', '2', '3'], _extras={'_extras': 'test', 'bool': True})
+        >>> cfg
+        {'name': 'foo', 'command': ['1', '2', '3'], '_extras': {'_extras': 'test', 'bool': True}}
+        >>> cfg['name']
+        'foo'
+        >>> cfg['_extras']
+        {'_extras': 'test', 'bool': True}
+        """  # TODO: move param description from configuration.rst
         self.name: str = name
         self.directory: str = directory
         self.code_report: bool = code_report
@@ -50,21 +62,22 @@ class ProjectConfiguration:
         self.fail_tag: str = fail_tag
         self.if_env_set: str = if_env_set
         self.children: Optional['Variations'] = None
-        self.extras: Dict[str, str] = extras if extras else {}
+        self._extras: Dict[str, str] = {}
         for key, value in kwargs.items():
-            self.extras[key] = value
+            self._extras[key] = value
 
     def __repr__(self) -> str:
         """
-        This function simulates dict-like legacy representation for output
+        This function simulates dict-like representation for string output. This is useful when printing contents of
+        :class:`.Variations` objects, as internally they wrap a list of :class:`.ProjectConfiguration` objects
         :return: dict-like string
 
         >>> cfg = ProjectConfiguration(name='foo', command=['bar'], my_var='baz')
         >>> repr(cfg)
         "{'name': 'foo', 'command': 'bar', 'my_var': 'baz'}"
         """
-        res = {k: v for k, v in self.__dict__.items() if v and k != 'extras'}
-        res.update(self.extras)
+        res = {k: v for k, v in self.__dict__.items() if v and k != '_extras'}
+        res.update(self._extras)
         if len(self.command) == 1:  # command should be printed as one string, instead of list
             res['command'] = self.command[0]
         return str(res)
@@ -91,6 +104,8 @@ class ProjectConfiguration:
         True
         >>> cfg1 == {'name': 'foo', 'my_var': 'bar', 'test': None}
         True
+        >>> cfg1 == {'name': 'foo', 'my_var': 'bar', 'test': ''}
+        True
         >>> cfg1 == {'name': 'foo', 'my_var': 'bar', 'test': ' '}
         False
         """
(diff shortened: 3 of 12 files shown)


universum/configuration_support.py
```diff
@@ -13,6 +16,47 @@
 skip_attributes = {"children", "critical", "skip_numbering_level"}
 
 
+ProjectConfigKeyType = Literal["name", "command", "environment", "artifacts", "report_artifacts",
+                               "artifact_prebuild_clean", "directory", "critical", "background", "finish_background",
+                               "code_report", "pass_tag", "fail_tag", "if_env_set"]
+
+
+class ProjectConfig(TypedDict, total=True):
```
i-keliukh, line 24: Why total=True?

universum/configuration_support.py
```diff
@@ -13,6 +16,47 @@
...
+    """
+    :param name: Specifies the name of the configuration.
+    :type name: str
+    The name is used as the title of the build log block corresponding to the build of this configuration. It is
+    also used to generate name of the log file if the option for storing to log files is selected. A project
+    configuration can have no name, but it is not recommended for aesthetic reasons. If several project
+    configurations have the same names, and logs are stored to files (see --launcher-output / -lo command-line
```
k-dovgan, line 31: ``--launcher-output`` / ``-lo`` still looks ok in docstrings, and might look way better for generated documentation

universum/configuration_support.py
```diff
@@ -13,6 +16,47 @@
...
+    parameter for details), all logs for such configurations will be stored to one file in order of their appearances.
+
+    :param command: The list with command line for the configuration launch.
+    :type command: list
```
k-dovgan, line 35: Shouldn't it be, like, 'list of str', or something alike? It must not be just a list of anything...

universum/configuration_support.py
```diff
@@ -13,6 +16,47 @@
...
+    For every project configuration first list item should be a console command (e.g. script name, or any other
+    command like ls), and other list items should be the arguments, added to the command (e.g. --debug or -la). Every
+    command line element, separated with space character, should be passed as a separate string argument. Lists like
+    ["ls -a"] will not be processed correctly and thus should be splat into ["ls", "-a"]. Lists like ["build.sh",
+    "--platform A"] will not be processed correctly and thus should be plat into ["build.sh", "--platform",
+    "A"]. A project configuration can have an empty list as a command. Such configuration won’t do anything except
+    informing user about missing it.
```
k-dovgan, line 42: Looks like you took the description from documentation; but it doesn't look nice now that I'm reading it. Can you, maybe, rephrase it a little to make it more understandable? Especially the thing about spaces and _arguments_
(You may even apply the same changes to actual documentation paragraph you took this description from in the same PR)

universum/configuration_support.py
```diff
@@ -13,6 +16,47 @@
...
+    """
```
k-dovgan, line 43: Will further descriptions be added here later?

universum/configuration_support.py
o-andrieiev: fixed.

universum/configuration_support.py
o-andrieiev: I don't follow this comment. Would you like me to change something?

universum/configuration_support.py
o-andrieiev: ok, I'll include these changes in the PR

universum/configuration_support.py
o-andrieiev: yep, I just submitted first take, since I wasn't sure whether I should continue with this work.
Having ProjectConfigKeyType and ProjectConfig keys as separate items is not ideal.
However this change does let us move all the information from configuring.html to configuration_support.html, removing the former page completely.

universum/configuration_support.py
k-dovgan: Yes, I meant changing '-quotes to `-quotes exactly for reasons demonstrated in comment above.

o-andrieiev: I just fully copied descriptions from 'configuring.rst' instead

universum/configuration_support.py
o-andrieiev: yep, I just submitted first take, since I wasn't sure whether I should continue with this work. Having ProjectConfigKeyType and ProjectConfig keys as separate items is not ideal. However this change does let us move all the information from configuring.html to configuration_support.html, removing the former page completely.

k-dovgan:
https://universum.readthedocs.io/en/type-annot/

universum/configuration_support.py
```diff
@@ -12,6 +15,203 @@
 
 skip_attributes = {"children", "critical", "skip_numbering_level"}
 
+ProjectConfigKeyType = Literal["name", "command", "environment", "artifacts", "report_artifacts",
+                               "artifact_prebuild_clean", "directory", "critical", "background", "finish_background",
+                               "code_report", "pass_tag", "fail_tag", "if_env_set"]
+
+
+class ProjectConfig(TypedDict, total=False):
```
i-keliukh, line 23: 1. Since users are not going to use this class, it has to be removed from the documentation.
Alternatively, we can change the approach to creating configs and let users to use this class.
However, this would be breaking change.
2. If we are discussing breaking changes, maybe it would be better to move away from a dictionary.
As an option, Variations can convert dictionary to the class automatically, but accept both variants.

universum/configuration_support.py
```diff
@@ -12,6 +15,203 @@
...
+    report_artifacts: List[str]
+    artifact_prebuild_clean: bool
+    directory: str
+    critical: bool
+    background: bool
+    finish_background: bool
+    code_report: str
+    pass_tag: str
+    fail_tag: str
+    if_env_set: str
```
i-keliukh, line 213: 1. Are you sure we are not limiting users to the listed key names?
Currently we are ignoring all unknown key names and, as fat as I know, some users already are using custom keys.
2. Can we avoid duplication between this list and ProjectConfigKeyType?

universum/configuration_support.py
```diff
@@ -12,6 +15,203 @@
 
 skip_attributes = {"children", "critical", "skip_numbering_level"}
 
+ProjectConfigKeyType = Literal["name", "command", "environment", "artifacts", "report_artifacts",
```
i-keliukh, line 18: The benefits of using this annotation are not clear to me.
There is only one usage in the code with only one key name checked.
My guess is that you are going to expand the usage in further updates.
Please confirm it.

universum/configuration_support.py
o-andrieiev: as temporary solution, I kept inheritance from dict, but I intend to remove that dependency in this PR, after I rewrite all other modules to use properties instead of accessing dict keys. I already updated Variations to accept either dict or ProjectConfiguration - which is the only change that would require new test case I think.

universum/configuration_support.py
o-andrieiev: I no longer think that we can remove configuring.html completely, but the "common keys" will move to configuration_support.html as part of ProjectConfiguration class description

tests/deployment_utils.py
```diff
@@ -128,7 +128,7 @@ def exit(self):
 
 
 @pytest.fixture()
-def execution_environment(request) -> ExecutionEnvironment:
+def execution_environment(request) -> Generator[ExecutionEnvironment, None, None]:
```
i-keliukh, line 131: 1. This is unrelated change. In general I am ok to mix annotation in different modules inside one commit, but configs change to class is too big change on its own.
2. pytest fixture are quite special beasts. I don't think we could benefit from this annotation anyway. Does mypy know that the function here and the parameter in test function are connected?

universum/configuration_support.py
```diff
@@ -317,57 +422,73 @@ class Variations(list):
 
     """
 
-    def __add__(self, other: List) -> 'Variations':
+    def __init__(self, lst: Optional[Union[List[Dict[str, Any]], List[ProjectConfiguration]]]):
+        #  lst can be List[Dict[str, Any]] to support legacy cases - should be removed after project migration
+        self.configs: List[ProjectConfiguration] = []  # aggregation is used instead of inheritance for type safety
```
i-keliukh, line 427: Why can't we just inherit from list?

tests/deployment_utils.py
o-andrieiev: mypy flagged these 2 lines as errors, when I checked the test files. I am fine with moving this change to separate PR.

universum/configuration_support.py
i-keliukh: Answer: there is no way to inherit from typed list

universum/configuration_support.py
```diff
@@ -117,57 +421,73 @@ class Variations(list):
...
         :return: iterable for all dictionary objects in :class:`.Variations` list
         """
-        for obj_a in self:
+        for obj_a in self.configs:
 
             obj_a_copy = copy.deepcopy(obj_a)
 
-            if "children" in obj_a_copy:
-                for obj_b in obj_a_copy["children"].all():
+            if obj_a_copy.children:
```
i-keliukh, line 489: Is it equivalent change?

universum/modules/code_report_collector.py
```diff
@@ -39,29 +32,18 @@ def set_code_report_directory(self, project_root: str) -> None:
         if not os.path.exists(self.report_path):
             os.makedirs(self.report_path)
 
-    def prepare_environment(self, project_configs: List[ProjectConfig]) -> Variations:
-        afterall_steps: List[ProjectConfig] = []
-        for item in project_configs:
-            if not item.get("code_report", False):
+    def prepare_environment(self, project_config_variations: Variations) -> Variations:
+        afterall_steps: List[ProjectConfiguration] = []
```
i-keliukh, line 36: Would it be safe to use Variations here?

universum/modules/launcher.py
```diff
@@ -350,30 +357,27 @@ def __init__(self, *args, **kwargs):
...
+        self.code_report_collector = self.code_report_collector_factory()
         self.include_patterns, self.exclude_patterns = get_match_patterns(self.settings.step_filter)
 
     @make_block("Processing project configs")
-    def process_project_configs(self):
+    def process_project_configs(self) -> configuration_support.Variations:
         config_path = utils.parse_path(self.settings.config_path, self.settings.project_root)
         configuration_support.set_project_root(self.settings.project_root)
-        config_globals = {}
+        config_globals: Dict[str, configuration_support.Variations] = {}
```
i-keliukh, line 367: I am not sure regarding technical consequences of this declaration, but config_globals is just a dictionary of all global variables of the configs module. Potentially it can contain variable of any type.

universum/configuration_support.py
o-andrieiev: it's not equivalent, but I think this change is warranted: previously configuration could lack some fields, and universum had lots of 'get' calls with provided default values - this was risky in case default was incorrectly set to 2 different values in 2 different parts of code. Instead ProjectConfiguration is constructed with all properties set, but all of them are 'falsy' by default. This means that there's an internal guarantee of returned type (no 'Optional'), which simplifies the code.

universum/modules/code_report_collector.py
o-andrieiev: in afterall_steps ?

universum/modules/launcher.py
o-andrieiev: sure, I can remove this part, and add explicit type cast, where Variations are retrieved from dict

universum/modules/code_report_collector.py
i-keliukh: yes

universum/configuration_support.py
i-keliukh: Ok, just to make it clear: are you saying that obj_a_copy is guaranteed to have "children" field?
If it is guaranteed by the static typing, I'm ok with the change.

universum/modules/code_report_collector.py
o-andrieiev: I am not sure I understand, we can either add Variations, or populate list of ProjectConfigurations and convert it at the end - there should be no difference really. I think current approach (convert in the end) has slightly lower overhead, since adding Variations has one more call to go through.

universum/configuration_support.py
o-andrieiev: children is a property so it's guaranteed, but it can return None which is falsy

universum/modules/code_report_collector.py
i-keliukh: It's about consistency (readability). I interpret Variations as a type used to store list of ProjectConfiguration instances, and the performance difference is minuscule.
BTW, probably we should also rename or alias Variations to something more meaningful.

o-andrieiev: sure, I'll change it to Variations. I think there were several other places where that would make sense.
I propose following naming options: ConfigurationList, ConfigurationHolder.

i-keliukh: ConfigurationList looks good to me

universum/modules/launcher.py
```diff
@@ -22,14 +23,14 @@
...
+def make_command(name: str) -> sh.Command:
     try:
         return sh.Command(name)
-    except sh.CommandNotFound as e :
+    except sh.CommandNotFound as e:
         raise CiException(f"No such file or command as '{name}'") from e
 
 
-def check_if_env_set(configuration):
+def check_if_env_set(configuration: configuration_support.ProjectConfiguration) -> bool:
```
o-andrieiev, line 33: move to ProjectConfiguration class?

universum/configuration_support.py
```diff
@@ -1,19 +1,286 @@
...
+                result = True
+            command_line = command_line + " " if command_line else "" + argument
+        self.command = [command_line]
+        return result
+
+
+DictType = TypeVar('DictType', bound=dict)
+
+
+def combine(dictionary_a: DictType, dictionary_b: DictType) -> DictType:
```
o-andrieiev, line 282: mark deprecated?

universum/configuration_support.py
```diff
@@ -1,19 +1,286 @@
...
+
+    def stringify_command(self) -> bool:
+        result = False
+        command_line = ""
+        for argument in self.command:
+            if " " in argument:
+                argument = "\"" + argument + "\""
+                result = True
+            command_line = command_line + " " if command_line else "" + argument
+        self.command = [command_line]
```
o-andrieiev, line 275: self.comman is List of strings, but this used to assign one string. Change to Union or report as list of 1 element?

universum/configuration_support.py
```diff
@@ -1,19 +1,286 @@
...
+        self.extras: Dict[str, str] = extras if extras else {}
+        for key, value in kwargs.items():
+            self.extras[key] = value
+
+    def replace_string(self, from_string: str, to_string: str) -> None:
+        self.command = [word.replace(from_string, to_string) for word in self.command]
+        self.artifacts = self.artifacts.replace(from_string, to_string)
+        self.report_artifacts = self.report_artifacts.replace(from_string, to_string)
+        self.directory = self.directory.replace(from_string, to_string)
+        # should we update extras? (user-defined)
```
o-andrieiev, line 236: should we replace path in user-defined variables?

universum/configuration_support.py
o-andrieiev: it was decided to move documentation changes in separate PR

universum/configuration_support.py
k-dovgan: I'd say 'yes'. As a user, I'd expect a 'variable-like' to behave as a variable always, not only in some cases.

universum/modules/launcher.py
k-dovgan: Maybe, but not in this PR, that's a separate change.

doc/configuring.rst
```diff
@@ -311,11 +311,12 @@ pass_tag, fail_tag
...
     a tag with spaces in TeamCity's web-interface. Every tag is added (if matching condition) after executing
     build step it is set in, not in the end of all run.
-    ``pass_tag`` and ``fail_tag`` can also be used when `multiplying build configurations`_, like this:
+    ``pass_tag`` and ``fail_tag`` can also be used when
+    `multiplying build configuration <configuring.html#multiplying-build-configurations>`__, like this:
 
     .. testsetup::
 
-        #!/usr/bin/env {python}
+        #!/usr/bin/env python3.7
```
o-andrieiev, line 319: this was unintentional

universum/configuration_support.py
```diff
@@ -1,18 +1,120 @@
...
+
+    def __repr__(self) -> str:
+        res = {k: v for k, v in self.__dict__.items() if v and k != 'extras'}
+        res.update(self.extras)
+        if len(self.command) == 1:  # command should be printed as one string, instead of list
```
o-andrieiev, line 68: this way we don't have to use Union[str, List[str]] for command attribute, but can still print ProjectConfiguration cleanly

universum/configuration_support.py
```diff
@@ -1,18 +1,120 @@
...
+            res['command'] = self.command[0]
+        return str(res)
+
+    def __eq__(self, other: Any) -> bool:
```
o-andrieiev, line 72: this is to preserve ability to compare Variations with list[dict] (old format)

universum/configuration_support.py
```diff
@@ -1,18 +1,120 @@
...
+        if isinstance(other, dict):
+            for key, val in other.items():
+                if getattr(self, key, None) != val:
+                    return False
+            return True
+        return super().__eq__(other)
+
+    def __getitem__(self, item: str) -> Optional[str]:
```
o-andrieiev, line 80: this is to simulate dict-like access for other clients

universum/configuration_support.py
```diff
@@ -1,18 +1,120 @@
...
+            background=self.background or other.background,
+            finish_background=self.finish_background or other.finish_background,
+            code_report=self.code_report or other.code_report,
+            pass_tag=self.pass_tag + other.pass_tag,
+            fail_tag=self.fail_tag + other.fail_tag,
+            if_env_set=self.if_env_set + other.if_env_set,
+            extras=combine(self.extras, other.extras)
+        )
+
+    def stringify_command(self) -> bool:
```
o-andrieiev, line 101: this is such a weird function honestly

universum/modules/code_report_collector.py
```diff
@@ -39,30 +32,19 @@ def set_code_report_directory(self, project_root: str) -> None:
...
+    def prepare_environment(self, project_config_variations: Variations) -> Variations:
+        afterall_steps: Variations = Variations()
+        for item in project_config_variations.configs:
+            if not item.code_report:
                 continue
 
-            self.set_code_report_directory(self.settings.project_root)
-            temp_filename: str = "${CODE_REPORT_FILE}"
-            name: str = utils.calculate_file_absolute_path(self.report_path, item.get("name", "")) + ".json"
+            self.set_code_report_directory(self.settings.project_root)  # why is this in a loop?
```
o-andrieiev, line 41: why is this in a loop?

tests/deployment_utils.py
```diff
@@ -4,11 +4,11 @@
 import os
 from pwd import getpwnam
 
+from typing import Generator
 import docker
 import py
 import pytest
 from requests.exceptions import ReadTimeout
-
```
k-dovgan, line 11: Why did you remove this line? It was separating absolute imports from related, that's what we do, like, everywhere in this project.

universum/configuration_support.py
o-andrieiev: fixed in altest update

tests/deployment_utils.py
o-andrieiev: by mistake, fixed now

universum/modules/vcs/perforce_vcs.py
```diff
@@ -510,7 +510,7 @@ def map_local_path_to_depot(self, report):
                 abs_path = self.p4.run("where", line["depotFile"])[0]["path"]
                 self.mappings_dict[abs_path] = line["depotFile"]
 
-    @make_block("Revert workspace to depot state")
+    @make_block("Revert workspace to depot state", pass_errors=False)
```
o-andrieiev, line 513: actually, I'm not sure why 'pass_errors=False' was missing here before

k-dovgan (dismissed):

miltolstoy (dismissed):

doc/configuring.rst
```diff
@@ -224,7 +224,7 @@ artifacts
     Path to the file or directory to be copied to the working directory as an execution result.
     Can contain shell-style pattern matching (e.g. `"out/*.html"`), including recursive wildcards
     (e.g. `"out/**/index.html"`).
-    If not stated otherwise (see ``--no-archive`` `command-line parameter <args.html#Artifact\ collection>`__
+    If not stated otherwise (see ``--no-archive`` `command-line parameter <args.html#Artifact\\ collection>`__
```
i-keliukh, line 227: The added slash break the link, see it here: https://universum.readthedocs.io/en/type-annot/configuring.html

k-dovgan (approved):

universum/modules/code_report_collector.py
i-keliukh: my guess it is "legacy reasons"
probably in original version there were no loop

i-keliukh: anyway, you can move it out of the loop

i-keliukh: actually, you should move it to after the loop: as I see now, the idea was not to create directory if there are no "code report" steps

universum/modules/code_report_collector.py
```diff
@@ -39,30 +32,19 @@ def set_code_report_directory(self, project_root: str) -> None:
...
+            temp_filename: str = "${CODE_REPORT_FILE}"  # TODO: move to global constant and update docs to use it
```
i-keliukh, line 42: What would be the benefit of using global constant?

universum/configuration_support.py
```diff
@@ -1,18 +1,210 @@
...
+        self.finish_background: bool = finish_background
+        self.pass_tag: str = pass_tag
+        self.fail_tag: str = fail_tag
+        self.if_env_set: str = if_env_set
+        self.children: Optional['Variations'] = None
+        self.extras: Dict[str, str] = extras if extras else {}
+        for key, value in kwargs.items():
+            self.extras[key] = value
+
+    def __repr__(self) -> str:
```
i-keliukh, line 57: Why we need custom `__repr__` function at all?
Isn't the default implementation for class enough?

o-andrieiev: default __repr__ returns "ProjectConfiguration object..."
Custom __repr__ helps with readable debug output + this way we can keep docstrings unchanged.

universum/configuration_support.py
```diff
@@ -1,18 +1,210 @@
...
+        :return: client-defined value
+
+        >>> cfg = ProjectConfiguration(name='foo', my_var='bar')
+        >>> cfg['name']
+        'foo'
+        >>> cfg['my_var']
+        'bar'
+        >>> cfg['test']
+        """
+        return self.extras.get(item, self.__dict__.get(item, None))
```
i-keliukh, line 120: From this implementation it looks like `self.extras` is kind of override for `self.__dict__`.
As can be seen in `self.__init__` there are two ways to add items into "self.extras": from explicitly passed "extras" argument and from keyword dictionary (kwargs). The first way allows to actually override "native" keys.
I my opinion there are several issues here:
 - do we need such overriding at all?
 - are we using "extras" argument anywhere?
 - "extras" work only for dict-like access (cfg['my_var']), but they don't work for object-like access (cfg.my_var)

universum/configuration_support.py
i-keliukh: What about other dict methods, such as `__len__`, `__setitem__`, `__delitem__`?
If we don't need them, we should mention this decision in the code.

universum/configuration_support.py
i-keliukh: Ok, I've got the debug output argument.
But what is the issue with docstrings exactly?

o-andrieiev: in docstrings there are many cases where indexed Variations object is printed. Since indexing Variations should return ProjectConfiguration - we need to redefine repr so that comparisong with dict contents works in docstring tests

universum/configuration_support.py
o-andrieiev: I assume ProjectConfigurations should be read-only. I'll define __setitem__ and __delitem__ to raise NotImplementedError

universum/configuration_support.py
o-andrieiev: "extras" exists to support client-defined dict values. With extras as it is - clients should notice no changes when this patch is applied. Internally universum should never use extras.
You're right that case when "extras" is a dict key deserves special attention - I'll add more tests for that case.

o-andrieiev: I don't think "extras" need to support object-like access, since they only exist to simulate legacy behaviour. I think best way would be to generate a warning anytime extras is used.

universum/configuration_support.py
i-keliukh: Ok, as I understand, previously it printed an array of dicts with their exact content. This is what the default `__repr__` for dict does. And since we are printing list, its `__str__` method uses `__repr__` of its elements for printing, not `__str__`. Unfortunately, the default `__repr__` for class is useless.

Please expand the explanation in the docstring.

o-andrieiev: exactly! I'll update the description

universum/configuration_support.py
```diff
@@ -1,18 +1,210 @@
...
+
+    def __add__(self, other: 'ProjectConfiguration') -> 'ProjectConfiguration':
+        """
+        This functions defines operator ``+`` for :class:`.ProjectConfiguration` class objects by
+        concatenating strings and contents of dictionaries
+
+        :param other: `ProjectConfiguration` object
+        :return: new `ProjectConfiguration` object, including all attributes from both `self` and `other` objects
```
i-keliukh, line 128: Please mention that "critical" attribute is skipped.

universum/configuration_support.py
```diff
@@ -117,7 +132,38 @@ def __getitem__(self, item: str) -> Optional[str]:
...
+        [<warnings.WarningMessage object at ...>]
+        >>> set_cfg_and_get_warnings(cfg, '_extras', 'baz')
+        []
+        >>> set_cfg_and_get_warnings(cfg, 'test', 42)
+        []
+        >>> cfg
+        {'name': 'bar', 'directory': 'foo', 'my_var': 'bar', '_extras': 'baz', 'test': 42}
+        """
+        if key in self.__dict__ and key != '_extras':
+            warn("Re-defining the value of a read-only ProjectConfiguration field", UserWarning, 3)
```
i-keliukh, line 163: Well, the field is not read-only, it just has to be accessed using object-like access.
I would give an example of proper usage in the warning, like the following:
"Re-defining the value of ProjectConfiguration field. Please use var.command to set it instead of using var['command']."

universum/configuration_support.py
```diff
@@ -33,8 +34,19 @@ def __init__(self,
...
                  fail_tag: str = '',
                  if_env_set: str = '',
-                 extras: Optional[Dict[str, str]] = None,
-                 **kwargs) -> None:  # explicit constructor for client's type safety
+                 **kwargs) -> None:
+        """
+        All params supplied via named attributed shall be  type-checked for safety. Clients may provide custom
+        parameters in kwargs - these parameters will be stored internally in '_extras' dict and can be retrieved by
+        indexing.
+        >>> cfg = ProjectConfiguration(name='foo', command=['1', '2', '3'], _extras={'_extras': 'test', 'bool': True})
```
i-keliukh, line 42: The doctest is confusing because you used the same name for the dictionary to hold extra values and for the extra value itself.

universum/configuration_support.py
o-andrieiev: defined setitem. Should delitem be defined still?

universum/configuration_support.py
o-andrieiev: fixed

universum/configuration_support.py
o-andrieiev: this was intentional to show that nothing breaks internally if names match

o-andrieiev: e.g. the clients/user doesn't need to care about _extras being reserved, etc

universum/configuration_support.py
o-andrieiev: should it really be skipped btw?

i-keliukh: The answer is in universum core logic.
The behavior of critical is different depending on where it is set to True.
If it is set in parent, but not in children, all children will be executed, but if any of them fails, the entire execution will stop after that. If critical is set on parent and children, the children will no be executed as soon as one of them fails.

Because of that, in contrast to all other attributes the propagation from parent to child does not happen.

i-keliukh (approved):

i-keliukh (dismissed):

universum/configuration_support.py
```diff
@@ -1,18 +1,265 @@
...
+
+        >>> cfg = ProjectConfiguration(name='foo', my_var='bar')
+        >>> cfg['name']
+        'foo'
+        >>> cfg['my_var']
+        'bar'
+        >>> cfg['test']
+        """
+        #  _extras are checked first - just in case _extras field is added manually
+        # do note that __setitem__ checks pre_defined fields first, however it's impossible to shadow them by
```
k-dovgan, line 138: predefined fields

i-keliukh (approved):

universum/configuration_support.py
```diff
@@ -1,18 +1,265 @@
...
+        self.if_env_set: str = if_env_set
+        self.children: Optional['Variations'] = None
+        self._extras: Dict[str, str] = {}
+        for key, value in kwargs.items():
+            self._extras[key] = value
+
+    def __repr__(self) -> str:
+        """
+        This function simulates dict-like representation for string output. This is useful when printing contents of
+        :class:`.Variations` objects, as internally they wrap a list of :class:`.ProjectConfiguration` objects
```
k-dovgan, line 74: As I understand it, :class:`Variations` without dot in the beginning should work as well, since this is the very same module. And it kinda looks better in my opinion.

i-keliukh (approved):

k-dovgan (approved):

PR #504, opened 2020-09-07
feat(args): print multiple errors in github

i-keliukh:
Moved read_and_check_multiline_option function from utils to
HasErrorState. Improved error messages text.

### DIFF
--- a/tests/test_multiline_variables.py
+++ b/tests/test_multiline_variables.py
@@ -3,47 +3,51 @@
 import os
 import subprocess
 
-import pytest
-
-from universum.lib.module_arguments import ModuleArgumentParser, IncorrectParameterError
-from universum.lib.utils import read_and_check_multiline_option
+from universum.lib.gravity import construct_component
+from universum.lib.module_arguments import ModuleArgumentParser
+from universum.modules.error_state import HasErrorState
 
 text = "This is text\nwith some line breaks\n"
 error_message = "This is some missing argument error message"
 
 script = f"""
 from universum.lib.module_arguments import ModuleArgumentParser
-from universum.lib.utils import read_and_check_multiline_option
+from universum.lib.gravity import construct_component
+from universum.modules.error_state import HasErrorState
 
 parser = ModuleArgumentParser()
-parser.add_argument('--argument', '-a', metavar='ARGUMENT')
+parser.add_argument('--argument', '-a', dest='HasErrorState.argument', metavar='ARGUMENT')
 
 namespace = parser.parse_args()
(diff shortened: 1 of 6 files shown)


i-keliukh:
Relates to #261.

tests/test_multiline_variables.py
```diff
@@ -3,47 +3,51 @@
...
 
 parser = ModuleArgumentParser()
-parser.add_argument('--argument', '-a', metavar='ARGUMENT')
+parser.add_argument('--argument', '-a', dest='HasErrorState.argument', metavar='ARGUMENT')
 
 namespace = parser.parse_args()
-print(read_and_check_multiline_option(namespace, 'argument', "{error_message}"))
+module = construct_component(HasErrorState, namespace)
+print(module.read_and_check_multiline_option('argument', "{error_message}"))
+if module.is_in_error_state():
```
k-dovgan, line 24: Should we, maybe, also check success scenarios?

tests/test_multiline_variables.py
```diff
@@ -3,47 +3,51 @@
...
+    print(module.global_error_state.get_errors()[0])
 """
 
 
-@pytest.fixture()
-def parser():
-    parser = ModuleArgumentParser()
-    parser.add_argument('--argument', '-a', metavar='ARGUMENT')
-    yield parser
+def construct_test_module(args):
```
k-dovgan, line 29: Why is function better than fixture?

tests/test_multiline_variables.py
```diff
@@ -66,6 +70,4 @@ def test_multiline_variable_stdin(tmp_path):
     del env['ARGUMENT']
 
     result = subprocess.run(["python3.7", script_path, "-a", "-"], **common_args, input="", check=False)
-    assert result.returncode != 0
-    assert "IncorrectParameterError" in result.stderr
-    assert error_message in result.stderr
+    assert error_message in result.stdout
```
k-dovgan, line 73: Why have we removed error type check?

tests/test_multiline_variables.py
i-keliukh: All tests in this file check error and success scenarios.

tests/test_multiline_variables.py
i-keliukh: Fixture is only needed when there is something to tear down.
Function is easier to navigate and debug in IDE .

tests/test_multiline_variables.py
i-keliukh: Previously run of a test script result in an exception.
The script is changed to print error explicitly and it doesn't raise the exception.

k-dovgan (approved):

PR #510, opened 2020-09-08
fix(test): warning produced during pytest run

i-keliukh:
During tests execution, the pytest produces the warning regarding future
change of the default format of the junit report. The solution is to
explicitly set the format. Fixed by setting the format to the new
version.

Also fixed the warning caused by the regular expressions.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+junit_family=xunit2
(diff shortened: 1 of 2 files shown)


pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+junit_family=xunit2
```
k-dovgan, line 2: Can you please explain this line in comment or commit message?

i-keliukh: Updated the PR description

k-dovgan (approved):

PR #514, opened 2020-09-10
feat(args): printing multiple errors in swarm, vcs

i-keliukh:
Fixed the check for error state in 'MainVcs’ module. If the check is implemented
for error state, as it was before, the code doesn’t try to construct the code
review module in case of any errors across the entire universum. As the result,
there is no check for errors in the parameters of the code review modules and
these errors are not reported to user.

Moved the check for code review and producing error from VCS drivers to the
'MainVcs’ module.

Fixed minor issue with unnecessary try..except block in 'Vcs’ module. In the
past the attempt to access missing vcs type parameter in module settings would
result in AttributeError exception. There were no other sources of this
exception in the code block. However, since there is an explicit check for the
vcs type setting, the try..except block is not needed anymore.

### DIFF
--- a/tests/test_argument_check.py
+++ b/tests/test_argument_check.py
@@ -308,7 +308,7 @@ def test_swarm_changelist_incorrect_format():
     assert_incorrect_parameter(settings, "changelist for unshelving is incorrect")
 
 
-def test_vcs_type_and_config_path():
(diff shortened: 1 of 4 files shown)


i-keliukh:
Relates to #261

k-dovgan (approved):

PR #518, opened 2020-09-11
feat(out): detect non-UTF8 environment and adapt output

k-dovgan:


universum/modules/output/terminal_based_output.py
```diff
@@ -28,6 +29,10 @@ class TerminalBasedOutput(BaseOutput):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.block_level = 0
+        if locale.getpreferredencoding() == "UTF-8":
+            self.unicode_acceptable = True
```
i-keliukh, line 33: In my opinion the intermediate variable is not needed.
You can directly initialize self.block_end here.

tests/test_regression.py
```diff
@@ -45,3 +45,23 @@ def test_p4_multiple_spaces_in_mappings(perforce_workspace, tmpdir):
     environment.settings.PerforceMainVcs.force_clean = True
     environment.settings.PerforceWithMappings.mappings = [f"{perforce_workspace.depot}   /..."]
     assert not __main__.run(environment.settings)
+
+
+def test_non_utf8_environment(docker_main):
+    output = docker_main.run("""
+from universum.configuration_support import Variations
```
i-keliukh, line 52: From the test code it looks like POSIX locale has UTF8 encoding (codeset) and en_US doesn't have it.
This is counterintuitive. 
If the test is correct, please provide explanations in source code comments.

i-keliukh (approved):

PR #520, opened 2020-09-15
feat(init): add a subcommand to create an example config

### LINKED ISSUE #517: Add 'init'/'create-config' functionality
Add subcommand, creating empty example project and printing a command line for its execution.
For example:
```bash
$ python3.7 -m universum init
Writing configs.py... [Done]

Type the following command to launch: python3.7 -m universum -vt=none -fsd=./ -lcp=./configs.py
```

k-dovgan:
Resolves #517


universum/initializer.py
```diff
@@ -0,0 +1,60 @@
...
+
+__all__ = ["Initializer"]
+
+
+class Initializer(Module):
+    description: str = "Create a dummy project"
+
+    @staticmethod
+    def define_arguments(parser: ModuleArgumentParser) -> None:
+        parser.add_argument('--target-direcotry', '-td', dest='target_dir', metavar="TARGET_DIRECTORY",
```
k-dovgan, line 15: Should we use arguments or interactive dialogue?

k-dovgan:
See documentation here: https://universum.readthedocs.io/en/517-universum-init/

universum/modules/output/output.py
```diff
@@ -70,3 +73,21 @@ class HasOutput(Module):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self.out: Output = self.out_factory()
+
+
+class MinimalOut:
+    def __init__(self, silent=False):
+        self.silent = silent
```
o-andrieiev, line 80: Please note that default behaviour of old MinimalOut used to be not to print anything (silent = True).
I think it's fine to have silent=False as default, just making sure the change of default behaviour is intentional.

doc/init.rst
```diff
@@ -0,0 +1,11 @@
+:orphan:
+
+Create dummy project
+--------------------
+
+After installing Universum via ``pip3.7 install -U universum`` or simply from a downloaded project root directory
```
o-andrieiev, line 6: "downloaded project root directory" might be a bit confusing, could you please rephrase?

universum/initializer.py
```diff
@@ -0,0 +1,51 @@
...
+echo "Replace this script with actual project sources"
+""".format(artifact_name))
+        script.chmod(0o777)
+
+        config = target_dir_path / config_name
+        config.write_text("""#!/usr/bin/env python3.7
+
+from universum.configuration_support import Variations
+
+configs = Variations([dict(name='Show directory contents', command=['ls', '-la']),
```
o-andrieiev, line 38: do we have a test for this?
I think we should have simple sanity check - otherwise if something changes in the 'configuration_support' module, we might forget to change this config to make sure it still works.

universum/modules/output/output.py
i-keliukh: I agree that silent=False by default is more natural than the previous behavior.

doc/init.rst
```diff
@@ -0,0 +1,11 @@
...
+run ``python3.7 -m universum init`` to create an example source directory with example config and get a command
+to run it.
+
+After doing so replace ``universum_dummy_project`` with actual project sources, and edit ``universum_config.py``
```
i-keliukh, line 10: Is comma missing here?

universum/initializer.py
```diff
@@ -0,0 +1,51 @@
+from pathlib import Path
+
+from .modules.output.output import MinimalOut
+from .lib.gravity import Module
+from .lib.module_arguments import ModuleArgumentParser
+
+__all__ = ["Initializer"]
+
+
+class Initializer(Module):
```
i-keliukh, line 10: I think "Initializer" is too abstract name.
Please change to "ProjectInitializer" or even "ConfigsPyCreator".

doc/init.rst
o-andrieiev: I'd just replace "After doing so" with "Afterwards"

universum/initializer.py
```diff
@@ -0,0 +1,51 @@
...
+    description: str = "Create a dummy project"
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.out = MinimalOut()
+
+    def execute(self) -> None:
+        target_dir = "universum_dummy_project"
```
i-keliukh, line 18: I thought we agreed to create file in current directory.

universum/initializer.py
```diff
@@ -0,0 +1,51 @@
...
+        target_dir_path = Path(target_dir)
+        script_name = "dummy_script.sh"
+        config_name = "universum_config.py"
+        artifact_name = "artifact_example.txt"
+        self.out.log("Creating a dummy project in " + target_dir)
+        target_dir_path.mkdir(parents=True, exist_ok=True)
+
+        script = target_dir_path / script_name
+        script.write_text("""#!/usr/bin/env bash
```
i-keliukh, line 27: Do we need to create script file at all?

universum/modules/output/output.py
k-dovgan: Yes, I've changed it because I also find it more natural.

universum/initializer.py
k-dovgan: I thought we agreed to create an example source directory in current directory, to be than replaced with actual source directory. In my opinion it's more obvious and is easier to clean afterwards.

universum/initializer.py
k-dovgan: As this 'init' function is to show the user how to possibly run a project with Universum, yes, I think an example of "where to put project sources and how to handle them" will be more useful, than simply creating a sole configuration file. We have lots of those in documentation.

universum/config_creator.py
```diff
@@ -0,0 +1,36 @@
...
+from universum.configuration_support import Variations
+
+configs = Variations([dict(name='Show directory contents', command=['ls', '-la']),
+                      dict(name='Print a line', command=['bash', '-c', 'echo Hello world'])])
+
+if __name__ == '__main__':
+    print(configs.dump())
+""")
+        self.out.log("To run with Universum, execute the following command:\n"
+                     "python3.7 -m universum nonci --launcher-config-path={} --artifact-dir={}"
```
i-keliukh, line 32: Why we are recommending to change the artifact directory setting from default?

tests/test_create_config.py
```diff
@@ -0,0 +1,13 @@
+from pathlib import Path
+import subprocess
+
+
+def test_create_config():
+    result = subprocess.run(["python3.7", "-m", "universum", "create-config"], capture_output=True, check=True)
+    new_command = result.stdout.splitlines()[3]
```
i-keliukh, line 7: This approach to extracting the command-line from the output is too rigid.
It creates the hidden dependency between the output and the test.
I propose to add some marker, like natural dollar sign before the printed command-line and base the extraction on the marker.

doc/args.rst
```diff
@@ -35,7 +35,8 @@ via ``python3.7 -m universum``.
         | * -f='test 1:!unit test 1'    - run all steps with 'test 1' substring in their names except those
          containing 'unit test 1'
 
-    {poll,submit,nonci,github-handler} : @replace
+    {create-config,poll,submit,nonci,github-handler} : @replace
+        | :doc:`universum init <create_config>`
```
i-keliukh, line 39: Is it correct reference?

doc/index.rst
```diff
@@ -28,3 +28,8 @@ The goal of this project is to provide unified approach for adding continuous in
 to any project. It currently supports Perforce, Git, Gerrit, Swarm, Jenkins and TeamCity.
 
 Sometimes `Universum` system can be referred to as the framework or just CI.
+
+To install Universum, make sure to :doc:`meet prerequisites <prerequisites>` and then simply run
+``pip3.7 install -U universum`` from command line.
+
+To :doc:`create an example config and execute it with Universum <init>`, run ``python3.7 -m universum init``.
```
i-keliukh, line 35: I am not sure what you meant by "execute" here.

universum/modules/output/output.py
```diff
@@ -1,12 +1,15 @@
-from typing import ClassVar, Union
+import sys
+from types import TracebackType
+from typing import ClassVar, Optional, Union
```
i-keliukh, line 3: I believe ClassVar and Union are not used

k-dovgan: Yes they are. ClassVar is in line 71 and Union is in line 29

i-keliukh: I see, they were used in non-modified code.

doc/index.rst
```diff
@@ -32,4 +32,4 @@ Sometimes `Universum` system can be referred to as the framework or just CI.
 To install Universum, make sure to :doc:`meet prerequisites <prerequisites>` and then simply run
 ``pip3.7 install -U universum`` from command line.
 
-To :doc:`create an example config and execute it with Universum <init>`, run ``python3.7 -m universum init``.
+To :doc:`create an example config and run it with Universum <init>`, run ``python3.7 -m universum init``.
```
i-keliukh, line 35: Does "python3.7 -m universum init" run the example config?

i-keliukh (approved):

PR #522, opened 2020-09-22
feat(launcher): add default config name

k-dovgan:
.

### DIFF
--- a/tests/test_argument_check.py
+++ b/tests/test_argument_check.py
@@ -188,11 +188,11 @@ def test_missing_params_correct_error(test_type, module, field, vcs_type, error_
     if test_type == "main":
         if module == "Vcs" and field == "type":
             return
-        elif module == "LocalMainVcs" and field == "source_dir":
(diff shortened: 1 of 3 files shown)


tests/test_argument_check.py
```diff
@@ -189,10 +186,19 @@ def test_missing_params_correct_error(test_type, module, field, vcs_type, error_
...
+        elif module == "GitVcs" and field == "refspec" and settings.Vcs.type == "gerrit":
+            return
+        elif module in ["PerforceVcs", "PerforceMainVcs"]:
+            if field == "port":
+                settings.PerforceVcs.user = ""
+            else:
+                settings.PerforceVcs.port = ""
         else:
-            settings.Launcher.config_path = ""
+            settings.Vcs.type = ""
```
i-keliukh, line 201: This line prevents vcs errors from happening.
Therefore there is no chance to check if vcs errors are produced even if vcs parameters are properly specified.

i-keliukh (approved):

PR #523, opened 2020-09-23
refactor(docs): add replacing {python} with actual Python version

k-dovgan:
.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -185,3 +185,25 @@ release = universum.__version__
 
 autodoc_member_order = 'bysource'
 add_module_names = False
+
+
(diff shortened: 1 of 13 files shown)


doc/conf.py
```diff
@@ -185,3 +185,25 @@
 
 autodoc_member_order = 'bysource'
 add_module_names = False
+
+
+# Add replacement for Python version
+import sys
+
+
+def ultimateReplace(app, docname, source):
```
i-keliukh, line 194: Should we add a link to stackoverflow?

i-keliukh (approved):

PR #524, opened 2020-09-24
refactor: use launched Python version instead of hardcoded

k-dovgan:


tests/test_api.py
```diff
@@ -1,16 +1,19 @@
...
 
+from .utils import PYTHON_VERSION
+
 config = """
 from universum.configuration_support import Variations
 
 configs = Variations([dict(name="Run script", artifacts="output.json",
-                           command=["bash", "-c", "python3.7 -m universum api file-diff > output.json"])])
-"""
+                           command=["bash", "-c", "python{} -m universum api file-diff > output.json"])])
```
i-keliukh, line 10: Please switch to f-string from `.format`

tests/test_code_report.py
```diff
@@ -21,7 +21,7 @@ def get_config(args: List[str]):
         from universum.configuration_support import Variations
 
         configs = Variations([dict(name="Run static pylint", code_report=True,
-            command=['python3.7', '-m', 'universum.analyzers.pylint'{''.join(args)}])])
+            command=['python{utils.PYTHON_VERSION}', '-m', 'universum.analyzers.pylint'{''.join(args)}])])
```
i-keliukh, line 24: I would prefer to use consistent approach:
1. always import python version variable into you module and refer to it without specifying utils module 
2. use the shorter name
3. include "python" into the variable

tests/test_code_report.py
```diff
@@ -82,26 +82,29 @@ def test_pylint_analyzer_wrong_params(runner_with_pylint, args, expected_log):
...
+    'bash', '-c', '""" + cmd + """'
 ])])"""
 
     env.configs_file.write(config.format(os.getcwd(), str(tmpdir)))
 
     res = __main__.run(env.settings)
 
     assert res == 0
-    stdout_checker.assert_has_calls_with_param(log_fail, is_regexp=True)
+    # stdout_checker.assert_has_calls_with_param(log_fail, is_regexp=True)
```
i-keliukh, line 109: Commented text

tests/test_multiline_variables.py
```diff
@@ -61,13 +62,13 @@ def test_multiline_variable_stdin(tmp_path):
     env['PYTHONPATH'] = os.getcwd()
     common_args = {"capture_output": True, "text": True, "env": env}
 
-    result = subprocess.run(["python3.7", script_path, "-a", "-"], **common_args, input=text, check=True)
+    result = subprocess.run(["python" + PYTHON_VERSION, script_path, "-a", "-"], **common_args, input=text, check=True)
```
i-keliukh, line 65: https://www.python.org/dev/peps/pep-0394 recommends to use `sys.executable`.
It is not suitable for tests in docker, but should work fine for host.
It is up to you to decide which way is better.

k-dovgan: The output of `sys.executable` is `/usr/bin/python3.7`. I personally don't like it and prefer `python3.7` everywhere.

tests/deployment_utils.py
```diff
@@ -227,16 +228,16 @@ def run(self, config: str, force_installed: bool = False, vcs_type: str = "none"
...
-        will not be used instead of those installed into system. Without '-I' option `python3.7 -m` will first
+        will not be used instead of those installed into system. Without '-I' option `python -m` will first
         try to launch universum from sources in `workdir` if there are any. That is why, if `workdir` is not
         default and there are no universum sources in specified `workdir`, the preinstalled universum will
         be ran as in case of `force_installed`.
         """
 
         if force_installed:
-            cmd = "python3.7 -I -m universum"
+            cmd = "{} -I -m universum".format(PYTHON)
```
i-keliukh, line 238: Please use f-string

i-keliukh (approved):

tests/test_code_report.py
```diff
@@ -89,13 +93,15 @@ def test_code_report_extended_arg_search(tmpdir, stdout_checker):
 
     tmpdir.join("source_file.py").write(source_code + '\n')
 
+    cmd = "cd \"{0}\" && " + python() + \
```
i-keliukh, line 96: What is the benefit of introducing this variable?

k-dovgan: Increased readability. Because of actual variables, inserted variables and a fake variable for a file this config turned into an unreadable mess.

i-keliukh: Please switch to f-string in both places

tests/test_code_report.py
```diff
@@ -1,12 +1,16 @@
...
+import sys
 from typing import List
 
 import pytest
 
 from universum import __main__
 from . import utils
+from .utils import python
+
+python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
```
i-keliukh, line 13: Please move python_version variable to utils

i-keliukh (dismissed):

i-keliukh (dismissed):

i-keliukh (approved):

PR #528, opened 2020-10-02
fix(jenkins_plugin): handle failed steps items array properly

### LINKED ISSUE #527: Not all failed steps handled by Jenkins plugin

miltolstoy:
Resolves #527


i-keliukh (changes requested):
Could you, please, add a regression test for this case?

miltolstoy:
> Could you, please, add a regression test for this case?

Coloring stopping function was missed in test HTML. HTML updated.

i-keliukh (approved):

PR #539, opened 2020-10-26
move param desc from .rst to .py

o-andrieiev:
Update docstrings


k-dovgan:
https://universum.readthedocs.io/en/docs_update/

universum/configuration_support.py
```diff
@@ -36,9 +38,74 @@ def __init__(self,
...
+            Usually set in conjunction with adding ``--result-file="${CODE_REPORT_FILE}"`` to 'command' arguments.
+            Analyzers currently provided by Universum are: ``pylint``, ``svace`` and ``uncrustify``
+            (see `code_report parameters <code_report.html>`__ for details).
+        :param pass_tag: A tag used to mark successful TemCity builds. This tag can be set independenty
+            of `fail_tag` value per each step. The value should be set to a strings without spaces as acceptable by
+            TeamCity as tags. Every tag is added (if matching condition) after executing build step it is set in,
+            not in the end of all run.
+        :param fail_tag: A tag used to mark failed TemCity builds. See `pass_tag` for details.
+
+        :Example of a simple Step construction:
```
i-keliukh, line 108: This line causes the issue with white space.

doc/configuration_support.rst
```diff
@@ -11,8 +11,7 @@
 .. automodule:: universum.configuration_support
     :members:
     :special-members:
-    :exclude-members: __weakref__
+    :exclude-members: __weakref__, __init__
 
     :mod:`universum.configuration_support`
```
k-dovgan, line 16: Please replace this line with non-header.

universum/configuration_support.py
```diff
@@ -8,14 +8,134 @@
...
 
+# pylint: disable-msg=line-too-long
 class Step:
     """
-    Step is a collection of configurable project-specific data needed for Universum operations.
+    Step is a collection of configurable project-specific data entries needed for Universum functionality. Individual
+    build steps are collected in a :class:`Configuration` object, which supports arithmetic operations for
+    combination and chaining.
+
+    Step keys
```
i-keliukh, line 24: Section titles are not allowed here

doc/configuring.rst
```diff
@@ -180,187 +180,6 @@ The example configuration file declares the following `Universum` steps:
 4. Copy resulting directory `"out"` to the artifact directory
 
 
-.. _keys:
```
k-dovgan, line 183: Please make sure there's no broken links to 'keys' left in this document. And in change log.

universum/configuration_support.py
```diff
@@ -8,14 +8,133 @@
     "Step",
     "Configuration",
     "Variations",
+    "combine",
```
k-dovgan, line 11: `__all__` refers to public API provided by module. I'd rather not add the 'combine' function as a publicly available, but also removed the legacy 'Variations' class. It would be still available for direct import, so it's not a problem; but not adding such things to `__all__` shows that we do not expect the user to use them.

universum/configuration_support.py
```diff
@@ -8,14 +8,133 @@
...
     "set_project_root",
     "get_project_root"
 ]
 
 
+# pylint: disable-msg=line-too-long
```
k-dovgan, line 17: Maybe move this to the beginning of the file, as we disable this rule for the whole module.

universum/configuration_support.py
```diff
@@ -8,14 +8,133 @@
...
 class Step:
     """
-    Step is a collection of configurable project-specific data needed for Universum operations.
+    Step is a collection of configurable project-specific data entries needed for Universum functionality. Individual
```
k-dovgan, line 20: Can we describe it in a more... plain way? Like, 'Step is a class to describe all parameters, passed to a single Universum launch' _(it's a bad description, it's just an example)._
You may leave this sentence somewhere as well, I'm just afraid, being totally correct, it still is more confusing than clarifying.

universum/configuration_support.py
```diff
@@ -8,14 +8,133 @@
...
+    build steps are collected in a :class:`Configuration` object, which supports arithmetic operations for
+    combination and chaining.
```
k-dovgan, line 22: Is 'chaining' an arithmetic operation? I thought it was 'combination and addition'? If 'chaining' is more correct, we should fix the 'configuring' paragraph name as well.

universum/configuration_support.py
```diff
@@ -8,14 +8,133 @@
...
+
+    **Step keys**
+
+    Parameters supplied via named attributes during construction are annotated and can be type-checked for safety.
+    Clients may also provide custom parameters in kwargs - these parameters will be stored internally in `dict` and can
```
k-dovgan, line 27: I'd rather use term 'users', because it's also used everywhere else in Universum documentation.
Also, I'd prefer `kwargs` instead of kwargs.
Also, I'd say 'in a `dict` field/attribute' instead of simple 'in `dict`'.
Also, I'd replace 'provide' with 'define'.

universum/configuration_support.py
```diff
@@ -8,14 +8,133 @@
...
+    be retrieved by indexing. The following list enumerates all the keys supported and used by Universum:
```
k-dovgan, line 28: 'all the keys supported in Universum configs by default' (for you say earlier that users can define their own keys as well, and it's also kind of supported by Universum)

universum/configuration_support.py
o-andrieiev: Agree with removal of Variations, but I think combine should be public - but maybe in utils.

universum/configuration_support.py
o-andrieiev: or I can enable it back after the class definition. Which one is preferable?

universum/configuration_support.py
o-andrieiev: chaining is not an arithmetic operation, but that's what really happens when you multiple configurations (via children).

universum/configuration_support.py
```diff
@@ -8,14 +8,133 @@
...
+        matching pattern are found.
+    report_artifacts
+        Path to the special artifacts for reporting (e.g. to Swarm). Unlike
+        :ref:`artifacts <build_artifacts>`, `report_artifacts` are not obligatory and their absence is not
+        considered a build failure. A directory cannot be stored as a separate artifact, so when using
+        ``--no-archive`` option, do not claim directories as `report_artifacts`. Please note that any file can be
+        put as `artifact`, `report_artifact`, or both. A file that is both in `artifacts` and `report_artifacts`,
+        will be mentioned in a report and will cause build failure when missing.
+    artifact_prebuild_clean
+        A flag to signal that directory should be clean prior to execution. By default,
```
i-keliukh, line 70: This flag signals that steps-specific artifacts must be cleaned before the build. This could be a single file, several files, shell-style pattern or directory.

universum/configuration_support.py
```diff
@@ -7,15 +7,132 @@
...
+    finish_background
+        A flag used to signal that the current step should be executed only after ongoing
+        background steps (if any) are finished.
+    code_report
+        A flag used to signal that the current step performs static or syntax analysis of the code.
+        Usually set in conjunction with adding ``--result-file="${CODE_REPORT_FILE}"`` to 'command' arguments.
+        Analyzers currently provided by Universum are: ``pylint``, ``svace`` and ``uncrustify``
+        (see `code_report parameters <code_report.html>`__ for details).
+    pass_tag
+        A tag used to mark successful TemCity builds. This tag can be set independenty
```
i-keliukh, line 98: "TemCity" -> "TeamCity"

universum/configuration_support.py
```diff
@@ -7,15 +7,132 @@
...
+        matching pattern are found.
+    report_artifacts
+        Path to the special artifacts for reporting (e.g. to Swarm). Unlike
+        :ref:`artifacts <build_artifacts>`, `report_artifacts` are not obligatory and their absence is not
+        considered a build failure. A directory cannot be stored as a separate artifact, so when using
+        ``--no-archive`` option, do not claim directories as `report_artifacts`. Please note that any file can be
+        put as `artifact`, `report_artifact`, or both. A file that is both in `artifacts` and `report_artifacts`,
+        will be mentioned in a report and will cause build failure when missing.
+    artifact_prebuild_clean
+        A flag to signal that directory must be cleaned before the build. This could be a single file, several files,
```
i-keliukh, line 68: Two sentences slightly contradict each other. I propose to change them like below:
A flag to signal that artifacts must be cleaned before the build. Cleaning of single files, directories and sets of files defined by shell-style patterns is supported.

universum/configuration_support.py
k-dovgan: In this particular case I'd choose disabling for the whole file, so if needed we can add more description to `Configuration` as well.

doc/configuration_support.rst
```diff
@@ -11,8 +11,4 @@
 .. automodule:: universum.configuration_support
```
k-dovgan, line 11: to line 8:
"See also: Please see" looks a little awkward. Could you rephrase it?

i-keliukh:
Please also fix sphinx-build warnings:

```
/home/user/Projects/Universum/doc/../universum/configuration_support.py:docstring of universum.configuration_support.Step:34: WARNING: undefined label: clean_artifacts (if the link has no caption the label must precede a section header)
/home/user/Projects/Universum/doc/../universum/configuration_support.py:docstring of universum.configuration_support.Step:44: WARNING: undefined label: build_artifacts (if the link has no caption the label must precede a section header)
/home/user/Projects/Universum/doc/../universum/configuration_support.py:docstring of universum.configuration_support.Step:64: WARNING: undefined label: background_step (if the link has no caption the label must precede a section header)
```

doc/configuring.rst
```diff
@@ -48,7 +48,7 @@ Both `name` and `command`, however, can be undefined. A build step without a nam
 a build step without a command will do nothing, but will still appear in log (and have a step number).
```
k-dovgan, line 49: to line 6: please add a link to 'Project configuration structures' when referencing configuration file.

doc/configuring.rst
```diff
@@ -48,7 +48,7 @@ Both `name` and `command`, however, can be undefined. A build step without a nam
...
 
 For storing such parameters :mod:`universum.configuration_support` provides a class :class:`Step`, that has
-:ref:`a list of various step parameters <keys>`. Build steps can be :ref:`added, multiplied <combining>` and
+a list of various step parameters. Build steps can be :ref:`added, multiplied <combining>` and
```
k-dovgan, line 52: I understand that `Step` itself is a link to key list, but please leave a more obvious "blue" link too for those not attentive enough.

doc/configuring.rst
```diff
@@ -12,7 +12,8 @@ file name or file path, use ``--config`` / ``-cfg`` `command-line parameter <arg
 or `CONFIG_PATH` environment variable.
 
 Internally the config file is processed by the :mod:`universum.launcher` module. The path is passed
```
k-dovgan, line 14: please fix :mod:`universum.launcher` here and in its page to refer to the page (`universum.modules.launcher`)

universum/configuration_support.py
```diff
@@ -7,15 +8,131 @@
...
-    "Variations",
     "set_project_root",
     "get_project_root"
 ]
 
 
 class Step:
     """
-    Step is a collection of configurable project-specific data needed for Universum operations.
+    Step is a collection of configurable project-specific data entries needed for Universum functionality and passed to
```
k-dovgan, line 18: An instance of this class defines a single launch of an external command by Universum. <maybe a link to `configuring.rst` here> It determines execution parameters and result handling.

universum/configuration_support.py
```diff
@@ -7,15 +8,131 @@
...
+    a single Universum launch. Individual build steps are collected in a :class:`Configuration` object, which supports
```
k-dovgan, line 19: Individual build steps are collected in a `Configuration` object, that allows to create linear and nested step structures.
_(I don't think this is a right place to explain 'combining and chaining' anyway)_

i-keliukh:
There are more warnings:
```
/home/user/Projects/Universum/doc/additional_commands.rst:40: WARNING: undefined label: clean_artifacts (if the link has no caption the label must precede a section header)
../CHANGELOG.rst:169: WARNING: undefined label: clean_artifacts (if the link has no caption the label must precede a section header)
../CHANGELOG.rst:196: WARNING: undefined label: build_environment (if the link has no caption the label must precede a section header)
../CHANGELOG.rst:282: WARNING: undefined label: finish_background (if the link has no caption the label must precede a section header)
../CHANGELOG.rst:427: WARNING: undefined label: tc_tags (if the link has no caption the label must precede a section header)
../CHANGELOG.rst:438: WARNING: undefined label: clean_artifacts (if the link has no caption the label must precede a section header)
../CHANGELOG.rst:619: WARNING: undefined label: critical_step (if the link has no caption the label must precede a section header)
../CHANGELOG.rst:877: WARNING: undefined label: report_artifacts (if the link has no caption the label must precede a section header)
```

universum/configuration_support.py
```diff
@@ -7,15 +8,131 @@
...
+    operations for combination and chaining.
+
+    **Step keys**
+
+    Parameters supplied via named attributes during construction are annotated and can be type-checked for safety.
+    Users may also define custom parameters in `kwargs` - these parameters will be stored internally in `dict` fields
```
k-dovgan, line 25: in a separate `dict` field (as I understand, there's only one). Also, maybe add the field name?

o-andrieiev: each kwarg passed get its own dict field

doc/configuring.rst
o-andrieiev: Please check whether the proposed change is satisfactory.

k-dovgan (approved):

i-keliukh (approved):

PR #550, opened 2020-11-19
feat(test): add testing for multiple python versions

k-dovgan:


i-keliukh:
"verstions"

Makefile
```diff
@@ -1,7 +1,18 @@
-TEST_TARGETS = pytest doctest
+PYTHON_VERSION :=
```
i-keliukh, line 1: I propose to change like this:
```Makefile
PYTHON_VERSION ?= $(shell python -c 'import sys; print("%d.%d"% sys.version_info[0:2])' )
```

i-keliukh:
Also, I propose to change from python version to python interpreter name, like we did in the documentation.

i-keliukh:
Please also create a separate PR that adds python version to image name

i-keliukh (approved):

PR #552, opened 2020-11-26
fix(test): create-config test interfering with actual config

k-dovgan:
The reason why `.universum.py` unexpectedly disappeared.

### DIFF
--- a/tests/test_create_config.py
+++ b/tests/test_create_config.py
@@ -6,15 +6,14 @@ from .utils import python
 
 
 def test_create_config(tmpdir):
-    my_env = os.environ.copy()
-    my_env["PYTHONPATH"] = os.getcwd()
+    test_environment = dict(os.environ, PYTHONPATH=os.getcwd())
     result = subprocess.run([python(), "-m", "universum", "init"],
-                            capture_output=True, check=True, cwd=tmpdir, env=my_env)
+                            capture_output=True, check=True, cwd=tmpdir, env=test_environment)
     new_command = ''
     for line in result.stdout.splitlines():
         if line.startswith(b'$ '):           # result.stdout is a byte string
             new_command = line.lstrip(b'$ ')
-    new_result = subprocess.run(new_command.split(), capture_output=True, check=True, cwd=tmpdir, env=my_env)
+    new_result = subprocess.run(new_command.split(), capture_output=True, check=True, cwd=tmpdir, env=test_environment)
     Path(tmpdir, ".universum.py").unlink()
     artifacts = Path(tmpdir, "artifacts")
     artifacts.joinpath("CONFIGS_DUMP.txt").unlink()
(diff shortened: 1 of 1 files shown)


i-keliukh (approved):

tests/test_create_config.py
```diff
@@ -1,18 +1,22 @@
...
+import os
 import subprocess
 
 from .utils import python
 
 
-def test_create_config():
-    result = subprocess.run([python(), "-m", "universum", "init"], capture_output=True, check=True)
+def test_create_config(tmpdir):
+    my_env = os.environ.copy()
```
i-keliukh, line 9: In my opinion, the following construct is cleaner:
```python
test_environment=dict(os.environ, PYTHONPATH=os.getcwd())
```
Also note the better variable name.

tests/test_create_config.py
```diff
@@ -1,18 +1,22 @@
...
+    result = subprocess.run([python(), "-m", "universum", "init"],
+                            capture_output=True, check=True, cwd=tmpdir, env=my_env)
     new_command = ''
     for line in result.stdout.splitlines():
         if line.startswith(b'$ '):           # result.stdout is a byte string
             new_command = line.lstrip(b'$ ')
-    new_result = subprocess.run(new_command.split(), capture_output=True, check=True)
-    Path(".universum.py").unlink()
-    artifacts = Path("artifacts")
+    new_result = subprocess.run(new_command.split(), capture_output=True, check=True, cwd=tmpdir, env=my_env)
```
i-keliukh, line 17: The parameters repeat twice.
Maybe it's worth saving them to dict and reusing the dict on each call.

tests/test_create_config.py
o-andrieiev: agree

tests/test_create_config.py
o-andrieiev: agree

tests/test_create_config.py
```diff
@@ -1,18 +1,21 @@
...
     for line in result.stdout.splitlines():
         if line.startswith(b'$ '):           # result.stdout is a byte string
             new_command = line.lstrip(b'$ ')
-    new_result = subprocess.run(new_command.split(), capture_output=True, check=True)
-    Path(".universum.py").unlink()
-    artifacts = Path("artifacts")
+    new_result = subprocess.run(new_command.split(), capture_output=True, check=True, cwd=tmpdir, env=test_environment)
+    Path(tmpdir, ".universum.py").unlink()
+    artifacts = Path(tmpdir, "artifacts")
     artifacts.joinpath("CONFIGS_DUMP.txt").unlink()
```
i-keliukh, line 19: Since moving to the temp directory, do we still need to clean artifacts after test?

i-keliukh (approved):

PR #566, opened 2021-02-15
fix(docs): pin Sphinx version to last successful

k-dovgan:


setup.py
```diff
@@ -16,7 +16,7 @@ def readme():
 
 vcs = p4 + github
 
-docs = ('sphinx', 'sphinx-argparse', 'sphinx_rtd_theme')  # This extra is required for RTD to generate documentation
+docs = ('sphinx<3.5.0', 'sphinx-argparse', 'sphinx_rtd_theme')  # This extra is required for RTD to generate documentation
```
i-keliukh, line 19: Wouldn't it be better to use specific version?

setup.py
```diff
@@ -16,7 +16,7 @@ def readme():
 
 vcs = p4 + github
 
-docs = ('sphinx', 'sphinx-argparse', 'sphinx_rtd_theme')  # This extra is required for RTD to generate documentation
+docs = ('sphinx=3.4.3', 'sphinx-argparse', 'sphinx_rtd_theme')  # This extra is required for RTD to generate documentation
```
i-keliukh, line 19: As far as I remember, it should be "==3.4.3"

k-dovgan: You are right: https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#id4

i-keliukh (approved):

PR #557, opened 2020-12-10
feat(docs): add support of collapsible sections

k-dovgan:

PR #569, opened 2021-02-26
feat(docs): apply 'collapsible' to pages of documentation

k-dovgan:
Applies #557


k-dovgan:
Example: https://universum.readthedocs.io/en/apply-collapsible/configuring.html#build-step

i-keliukh:
In my opinion the collapsible block should be applied only if there are more than 4 lines of code.
BTW, now I think we should remove the animation.

k-dovgan:
> BTW, now I think we should remove the animation.

Any arguments why? I think it looks pretty and not annoying.

k-dovgan:
This is a test comment.

k-dovgan:
This is another test comment

k-dovgan:
This is one more test comment

i-keliukh (approved):

PR #578, opened 2021-04-01
fix(pylint): newer Pylint added separate setting for class constants

k-dovgan:


k-dovgan:
Pylint test ouside of venv was due to legacy reasons and is not needed any more.

i-keliukh (approved):

PR #583, opened 2021-04-15
feat(analyzers): add mypy support

o-andrieiev:


universum/analyzers/mypy.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Mypy runner.
+    Specify parameters such as project folders, config file for code report tool.
+    For example:
+    ./mypy.py --python-version 2 --files *.py tests/
+    Output: json of the found issues in the code.
+    """
+
+    @staticmethod
+    def define_arguments():
```
k-dovgan, line 19: The arguments here are almost exactly the same, as module 'pylint' arguments. Have you considered moving them to 'utils'? E.g. to avoid almost identical, yet quite long explanation of what '--python-version' is

o-andrieiev: yes, I did. Also, "--files" argument is common with pylint and uncrustify. However, I consider refactoring of analyzers a separate item from adding mypy. I think they should be done as distinct PRs

o-andrieiev: My priorities for now are:
1. Check that mypy handler correctly adds comments for review (not sure how to reproduce that at the moment)
2. Expand test suite for analyzers within this PR - right now it's very limited
3. Refactor analyzers as a separate PR

tests/docker/perforce/Dockerfile
```diff
@@ -6,11 +6,11 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get install -y wget gnupg2
 
 # Get perforce packages
-RUN wget -q http://package.perforce.com/perforce.pubkey -O - | apt-key add - && \
-    echo "deb http://package.perforce.com/apt/ubuntu bionic release" > /etc/apt/sources.list.d/perforce.list && \
-    apt update
+#RUN wget -q http://package.perforce.com/perforce.pubkey -O - | apt-key add - && \
```
k-dovgan, line 9: What is the purpose of these changes?

o-andrieiev: oops, these were not supposed to be committed

i-keliukh (dismissed):

k-dovgan (dismissed):
Looks awesome

universum/analyzers/utils.py
```diff
@@ -2,13 +2,28 @@
...
+
+
+def add_result_file_argument(parser):
     parser.add_argument("--result-file", dest="result_file",
                         help="File for storing json results of Universum run. Set it to \"${CODE_REPORT_FILE}\" "
                              "for running from Universum, variable will be handled during run. If you run this "
                              "script separately from Universum, just name the result file or leave it empty.")
 
 
+def add_python_version_argument(parser):
```
k-dovgan, line 19: But we have other analyzers then pylint and mypy. Uncrusify doesn't need 'python version'

tests/test_code_report.py
```diff
@@ -10,49 +10,64 @@
...
     docker_main.environment.install_python_module("pylint")
+    docker_main.environment.install_python_module("mypy")
     yield docker_main
 
 
-def get_config(args: List[str]):
-    args = [f", '{arg}'" for arg in args]
-    return inspect.cleandoc(f"""
-        from universum.configuration_support import Configuration
+class ConfigData:
```
o-andrieiev, line 20: class added to simplify construction of config with multiple analyzers

.universum.py
```diff
@@ -2,7 +2,7 @@
 
 import os
 import sys
-from universum.configuration_support import Variations
+from universum.configuration_support import Variations, Step
```
o-andrieiev, line 5: This change might not be fitting in this PR, but it's fairly small and simple. It seems I missed some internal usages of old config format.

.universum.py
```diff
@@ -17,36 +17,38 @@ def pip_install(module_name):
...
+                      Step(name="Run static pylint", code_report=True,
                            command=run_virtual(f"{python} -m universum.analyzers.pylint "
                                                f"--python-version={python_version} --rcfile=pylintrc "
                                                "--result-file=\"${CODE_REPORT_FILE}\" --files *.py universum/ tests/")),
-                      dict(name="Run static type checker", code_report=True,
-                           command=run_virtual("make mypy")),
+                      Step(name="Run static type checker", code_report=True,
+                           command=run_virtual(f"{python} -m universum.analyzers.mypy "
+                                               f"--python-version={python_version} "
+                                               "--result-file=\"${CODE_REPORT_FILE}\" --files *.py universum/")),
```
o-andrieiev, line 37: note that "tests" are not in scope

tests/test_code_report.py
```diff
@@ -10,49 +10,64 @@
...
+    for analyzer in analyzers:
+        config.add_analyzer(analyzer, common_args)
+
+    log = runner_with_analyzers.run(config.finalize())
+    if expected_success:
+        assert re.findall(log_success, log)
+    else:
+        assert re.findall(log_fail, log)
+        for analyzer in analyzers:
+            assert re.findall(fr'Run {analyzer} - [^\n]*Failed', log)
```
o-andrieiev, line 66: to make sure that second analyzer is checked even if first failed

tests/test_code_report.py
```diff
@@ -61,25 +76,32 @@ def test_without_code_report_command(runner_with_pylint):
     assert not pattern.findall(log)
 
 
-@pytest.mark.parametrize('args, expected_log', [
-    [["--python-version", python_version(), "--files", "source_file.py", "--result-file", "${CODE_REPORT_FILE}", '--rcfile'],
-     'rcfile: expected one argument'],
+@pytest.mark.parametrize('analyzer', ['pylint', 'mypy'])
```
o-andrieiev, line 79: pylint and mypy share a lot of input parameters, so it makes sense to group them like this

tests/test_code_report.py
```diff
@@ -61,25 +76,32 @@ def test_without_code_report_command(runner_with_pylint):
...
-
     [["--python-version", "--files", "source_file.py", "--result-file", "${CODE_REPORT_FILE}"],
      "python-version: expected one argument"],
     [["--python-version", python_version(), "--result-file", "${CODE_REPORT_FILE}"],
      "error: the following arguments are required: --files"],
 ])
-def test_pylint_analyzer_wrong_params(runner_with_pylint, args, expected_log):
-    source_file = runner_with_pylint.local.root_directory.join("source_file.py")
+def test_pylint_analyzer_wrong_common_params(runner_with_analyzers, analyzer, common_arg_set, expected_log):
+    test_pylint_analyzer_wrong_specific_params(runner_with_analyzers, analyzer, common_arg_set, expected_log)
```
o-andrieiev, line 91: I couldn't find a way to do "AxB + C" with pytest fixtures, so went with a bit hacky way

universum/analyzers/mypy.py
```diff
@@ -0,0 +1,73 @@
...
+        parser.add_argument("--config-file", dest="config_file", type=str, help="Specify a configuration file.")
+        utils.add_python_version_argument(parser)
+        utils.add_result_file_argument(parser)
+        return parser
+
+    def __init__(self, settings) -> None:
+        self.settings = settings
+
+    def execute(self) -> int:
+        cmd = [f"python{self.settings.version}", '-m', 'mypy', '--ignore-missing-imports']
```
o-andrieiev, line 35: most notably 'sh' lacks type annotations and there is no stub library

tests/test_code_report.py
```diff
@@ -10,49 +10,64 @@
...
-
-
-def test_without_code_report_command(runner_with_pylint):
-    log = runner_with_pylint.run("""
+def test_code_report_log(runner_with_analyzers, analyzers, common_args, tested_content, expected_success):
+    common_args.extend(["--python-version", python_version(), "--files", "source_file.py"])
+    runner_with_analyzers.local.root_directory.join("source_file.py").write(tested_content)
+    config = ConfigData()
+    for analyzer in analyzers:
+        config.add_analyzer(analyzer, common_args)
```
i-keliukh, line 58: Considering the common_args is the same for all test instances, please remove it from parametrization.

tests/test_code_report.py
i-keliukh: Maybe add this as a comment to source code?

tests/test_code_report.py
i-keliukh: I am not sure pytest allows to call tests as regular functions.
As far as I know, we don't do that.

universum/analyzers/pylint.py
```diff
@@ -20,17 +20,10 @@ class PylintAnalyzer:
     @staticmethod
     def define_arguments():
         parser = argparse.ArgumentParser(description="Pylint analyzer")
-        parser.add_argument("--files", dest="file_list", nargs='+', required=True,
-                            help="Python files and Python packages for Pylint. "
-                                 "Files could be defined as a single python file"
-                                 " *.py or directories with __init__.py file in the directory.")
+        utils.add_files_argument(parser)
```
i-keliukh, line 23: In my opinion you can merge all three calls into one.

universum/analyzers/utils.py
```diff
@@ -2,13 +2,28 @@
 import sys
 
 
-def add_common_arguments(parser):
+def add_files_argument(parser):
```
i-keliukh, line 5: If it is only for python files, the function name must reflect it.

o-andrieiev: No, the descripion is wrong (copied from pylint). Files argument is common for analyzers (e.g. uncrustify has it too)
However, I just noticed that destination for uncrustify's --files argument is file_names instead of file_list. I'll change the destination variable name

universum/analyzers/pylint.py
o-andrieiev: you mean like utils.add_files_python_version_and_result_file(parser) ?
that seems a bit excessive to me

universum/analyzers/utils.py
o-andrieiev: should be fixed in new revision

tests/test_code_report.py
o-andrieiev: should be fixed in new revision. I left it there in case we need args for python version check, but that stays as TODO for now

tests/test_code_report.py
o-andrieiev: done

tests/test_code_report.py
o-andrieiev: it seems to work, alternatively I can move the code to separate function run_with_params and call that twice

universum/analyzers/utils.py
```diff
@@ -1,32 +1,64 @@
...
-def add_files_argument(parser):
-    parser.add_argument("--files", dest="file_list", nargs='+', required=True,
-                        help="Python files and Python packages for analyzer. "
-                             "Files could be defined as a single python file"
-                             " *.py or directories with __init__.py file in the directory.")
+ReportData = TypedDict('ReportData', {'path': str, 'message': str, 'symbol': str, 'line': int})
 
 
-def add_result_file_argument(parser):
+def run_parse_report(cmd: List[str], parse: Callable[[str], List[ReportData]], out_file: str) -> int:
```
i-keliukh, line 13: The name makes an impression that the function runs a report.
I suggest to rename it like "run_and_parse_report".
I also suggest to rename "parse" argument to "parse_function".

tests/docker/universum_test_env/Dockerfile
```diff
@@ -4,6 +4,15 @@ ARG PYTHON=python3.7
...
 
+# Install python 2 for pylint testing
+RUN DEBIAN_FRONTEND=noninteractive apt install -y python
+RUN wget --no-check-certificate -O get-pip.py 'https://bootstrap.pypa.io/pip/2.7/get-pip.py'
+RUN python get-pip.py
+RUN python -m pip install pylint
+
+# Install uncrustify for testing
+RUN DEBIAN_FRONTEND=noninteractive apt install -y uncrustify
+
```
k-dovgan, line 15: Please move this block to the end of the file to use existing python layers (that are common for several different docker images)

o-andrieiev: moving python2 installation after python3 installation causes some issues

o-andrieiev:
Separated into multiple PRs to preserve change history and atomicity

PR #586, opened 2021-04-20
fix(p4): 'p4 opened' sometimes fails unexpectedly

k-dovgan:


universum/modules/vcs/perforce_vcs.py
```diff
@@ -476,7 +476,12 @@ def check_diff_for_depot(self, depot: str) -> str:
 
     def calculate_file_diff(self) -> List[Dict[str, str]]:
         action_list: Dict[str, str] = {}
-        for entry in self.p4.run_opened():
+        try:
```
i-keliukh, line 479: Please implement the following changes:

1. Logs of the operation
2. Retries with exponential increase in delay
3. If possible, process only one specific exception

universum/modules/vcs/perforce_vcs.py
```diff
@@ -476,7 +476,12 @@ def check_diff_for_depot(self, depot: str) -> str:
...
+            opened_files = self.p4.run_opened()
+        except P4Exception:
+            return [{'failure': "Getting diff failed due to server error!"}]
```
i-keliukh, line 482: In case of fatal failure, for example, if failure persists even after all retries, return empty list.
Do not return fake data to clients.

universum/modules/api_support.py
```diff
@@ -45,5 +46,10 @@ def get_environment_settings(self) -> Dict[str, str]:
     def add_file_diff(self, entry: str) -> None:
         self._set_entry("DIFF", entry)
 
+    def register_file_diff_failure(self) -> None:
```
i-keliukh, line 49: Considering there is only one symmetric function (get_file_diff), it is better to keep setting the value and error status within single function.

universum/modules/vcs/vcs.py
```diff
@@ -177,9 +177,15 @@ def prepare_repository(self) -> None:
         status_file.write("\nFile list:\n\n")
         status_file.write(utils.trim_and_convert_to_unicode(sh.ls("-lR", self.settings.project_root)) + "\n")
         status_file.close()
+        self.calculate_diff_for_api()
 
+    @make_block("Registering file diff for API")
+    def calculate_diff_for_api(self) -> None:
         file_diff = self.driver.calculate_file_diff()
-        self.api_support.add_file_diff(json.dumps(file_diff, indent=4))
+        if file_diff is False:
```
i-keliukh, line 185: I suggest to use None instead of False

i-keliukh (dismissed):

o-andrieiev (commented):
To fix static analysis issue, update P4 stubs as follows:

class P4ExceptionStub(BaseException):
    warnings: List[str]


class P4stub(ModuleType):  # replace with proper stubs
    P4Exception: Type[P4ExceptionStub]
    P4: Callable


P4Exception = P4ExceptionStub




...this can be done as a separate PR

universum/modules/vcs/perforce_vcs.py
```diff
@@ -474,9 +475,23 @@ def check_diff_for_depot(self, depot: str) -> str:
             result = utils.trim_and_convert_to_unicode(e.stdout)
         return result
 
-    def calculate_file_diff(self) -> List[Dict[str, str]]:
+    def calculate_file_diff(self) -> Union[List[Dict[str, str]], None]:
```
i-keliukh, line 478: Please change Union[something, None] to Optional[something]

k-dovgan:
> ...this can be done as a separate PR

Looks like a minor change to me, applied here.

tests/test_regression.py
```diff
@@ -110,3 +113,20 @@ def mocking_function(*args, **kwargs):
...
+    config = f"""
+from universum.configuration_support import Configuration
+
+configs = Configuration([dict(name="{step_name}", artifacts="output.json",
+                              command=["bash", "-c", "{python()} -m universum api file-diff > output.json"])])
+    """
+    settings = shelve_config(config, perforce_environment)
+
+    assert not __main__.run(settings)
+    log = perforce_environment.temp_dir.join('artifacts', f'{step_name}_log.txt').read()
```
i-keliukh, line 130: Please create dedicated `artifact_dir` variable in TestEnvironment and use it here.

i-keliukh (approved):

PR #591, opened 2021-04-28
fix(test): wrong workdir contained no sources

k-dovgan:
Also make sure reinstalled modules are updated even if not deleted


i-keliukh:
Please describe either in commit message or in source code comments what is the issue, how it manifests itself, how you found it, why we didn't see it before.

tests/test_regression.py
```diff
@@ -131,3 +132,28 @@ def test_p4_api_failed_opened(perforce_environment, mock_opened):
...
+
+
+def test_which_universum_is_tested(docker_main):
+    config = """
+from universum.configuration_support import Step, Configuration
+
+configs = Configuration([Step(name="Check python", command=["ls", "-la"])])
+"""
+    # THIS TEST PATCHES ACTUAL SOURCES!! BE CAREFUL
+    init_file = Path(docker_main.environment.get_working_directory()).joinpath("universum", "__init__.py")
```
i-keliukh, line 144: The path must be calculated based on the location of sources under test, not on the container parameters.

tests/test_regression.py
```diff
@@ -131,3 +132,28 @@ def test_p4_api_failed_opened(perforce_environment, mock_opened):
...
+    backup = init_file.read_bytes()
+    test_line = "THIS IS A TESTING VERSION"
```
i-keliukh, line 146: Please make the generated string random.

tests/test_regression.py
```diff
@@ -131,3 +132,28 @@ def test_p4_api_failed_opened(perforce_environment, mock_opened):
...
     assert "Getting file diff failed due to Perforce server internal error" in log
+
+
+def test_which_universum_is_tested(docker_main, pytestconfig):
+    config = """
+from universum.configuration_support import Step, Configuration
+
+configs = Configuration([Step(name="Check python", command=["ls", "-la"])])
+"""
+    # THIS TEST PATCHES ACTUAL SOURCES!! BE CAREFUL
```
i-keliukh, line 143: You should either put three exclamation marks or leave just one.
In my opinion, it would be better to tone down this comment.

i-keliukh (approved):

PR #596, opened 2021-05-05
refactor(analyzer): standardize & test uncrustify

o-andrieiev:


universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
+            line=line
+        ))
+
+    return result
+
+
+def _get_issue_message(before: str, after: str) -> str:
+    # The maximum number of lines to write separate comments for
+    # If exceeded, summarized comment will be provided instead
+    max_lines = 11
```
o-andrieiev, line 164: moved inside from global as this value only makes sense within the context of the function

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
+        sys.stderr.write("Please specify the '--cfg_file' parameter "
+                         "or set an env. variable 'UNCRUSTIFY_CONFIG'")
+        return 2
+    wrapcolumn, tabsize = _get_wrapcolumn_tabsize(settings.cfg_file)
+    html_diff_file_writer = HtmlDiffFileWriter(target_folder, wrapcolumn, tabsize)
+
+    files: List[Tuple[Path, Path]] = []
+    try:
+        src_files: List[Path] = _get_src_files(settings.file_list, settings.file_list_config)
+        for pattern in settings.pattern_form:
```
o-andrieiev, line 44: pattern matching used to be part of "parse_files" routine, but I think semantically populating the list and filtering the list should be separate functions

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
-        return file_issues
-
-    def execute(self):
-        if not self.settings.cfg_file and ('UNCRUSTIFY_CONFIG' not in os.environ):
-            sys.stderr.write("Please specify the '--cfg_file' parameter "
-                             "or set an env. variable 'UNCRUSTIFY_CONFIG'")
-            return 2
+        issues = _get_issues_from_diff(src_file, src_lines, fixed_lines)
+        if issues:
+            diff_file_writer.write_file(src_file, src_lines, fixed_lines)
```
o-andrieiev, line 123: writing html diff file used to be a part of get_file_issues function, but it seems to me that dependency on html is not needed here. In my opinion, parameter data such as target directory and format details for diff files does not belong to this function. Moved to separate 'writer' class, which is built ahead of time in main().

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
+        result.extend(issues)
+    return result
+
+
+def _get_wrapcolumn_tabsize(cfg_file: str) -> Tuple[int, int]:
```
o-andrieiev, line 128: this function is only called once now to build html diff file writer

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
-    return UncrustifyAnalyzer.define_arguments()
+def _get_text_for_block(start: int, end: int, lines: List[str]) -> str:
+    return _replace_invisible_symbols(''.join(lines[start: end]))
 
 
-def main():
-    analyzer_namespace = UncrustifyAnalyzer.define_arguments().parse_args()
-    analyze = UncrustifyAnalyzer(analyzer_namespace)
-    return analyze.execute()
+def _replace_invisible_symbols(line: str) -> str:
```
o-andrieiev, line 195: this function used to have a bug, where line was returned immediately after first replace

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
-            cmd("-c", self.settings.cfg_file, "--prefix", self.settings.output_directory, files)
-        except sh.ErrorReturnCode as e:
-            sys.stderr.write(str(e) + '\n')
 
-        issues_loads = []
-        for file_name in files:
-            issues_loads.extend(self.get_file_issues(file_name))
-        utils.output_report_to_file(issues_loads, self.settings.result_file)
-        return 1 if issues_loads else 0
+def _get_mismatching_block(previous_match: difflib.Match,  # src[a:a+size] = target[b:b+size]
```
o-andrieiev, line 175: changed match naming to better align with calling function's logic

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
+    previous_match_end_in_src = previous_match.a + previous_match.size
+    previous_match_end_in_target = previous_match.b + previous_match.size
+    match_start_in_src = match.a
+    match_start_in_target = match.b
+    if previous_match_end_in_src == match_start_in_src:
+        return None
+    line = match_start_in_src
+    before = _get_text_for_block(previous_match_end_in_src - 1, match_start_in_src, src)
+    after = _get_text_for_block(previous_match_end_in_target - 1, match_start_in_target, target)
+    return line, before, after
```
o-andrieiev, line 188: replaced dict with namedtuple to preserve type data

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
-        file_name = os.path.relpath(file_name, self.settings.output_directory).replace('/', '_') + '.html'
-        with open(os.path.join(self.settings.output_directory, file_name), 'w') as outfile:
-            outfile.write(differ.make_file(left_lines, right_lines, context=False))
-
-    def get_file_issues(self, src_file):
-        self.settings.output_directory = os.path.join(os.getcwd(), self.settings.output_directory)
-        # Uncrustify copies absolute path in its target folder, that's why we use '+'
-        uncrustify_file = os.path.normpath(self.settings.output_directory + '/' + src_file)
-
+def _uncrustify_output_parser(files: List[Tuple[Path, Path]],
```
o-andrieiev, line 111: uncharacteristically, this analyzer ignores std output and instead uses result files to determine the issues - hence "_" parameter

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
+            raise EnvironmentError("Please provide at least one file for analysis")
+    except EnvironmentError as e:
+        sys.stderr.write(str(e))
+        return 2
+
+    cmd = ["uncrustify", "-q", "-c", settings.cfg_file, "--prefix", settings.output_directory]
+    cmd.extend([str(path.relative_to(Path.cwd())) for path in src_files])
+
+    return utils.report_parsed_outcome(cmd,
+                                       lambda _: _uncrustify_output_parser(files, html_diff_file_writer, _),
```
o-andrieiev, line 61: lambda is used to pass additional settings-based configuration data without modifying the function signature in 'utils' and other analyzers

k-dovgan (approved):

tests/test_code_report.py
```diff
@@ -39,12 +40,24 @@ def finalize(self) -> str:
...
+code_width = 120
+input_tab_size = 2
+"""
 
 log_fail = r'Found [0-9]+ issues'
 log_success = r'Issues not found'
 
 
 @pytest.mark.parametrize('analyzers, extra_args, tested_content, expected_success', [
+    [['uncrustify'], [], source_code_c, True],
```
i-keliukh, line 59: Please assign human-readable test id to each test instance in this or separate PR.

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,196 @@
...
+            raise EnvironmentError("Please provide at least one file for analysis")
+    except EnvironmentError as e:
+        sys.stderr.write(str(e))
+        return 2
+
+    cmd = ["uncrustify", "-q", "-c", settings.cfg_file, "--prefix", settings.output_directory]
+    cmd.extend([str(path.relative_to(Path.cwd())) for path in src_files])
+
+    return utils.report_parsed_outcome(cmd,
+                                       lambda _: _uncrustify_output_parser(files, write_html_diff_file),
```
i-keliukh, line 54: Both the "write_html_diff_file" function and "_uncrustify_output_parser" function are used in this only place.
However, the former is local to the main function and is passed to the latter as a callback parameter.
As I see it, the only (technical) reason for this is using of "target_folder" and "differ" variables.
At the same time the lambda uses "files" variable.

Please consider to uniform the usage of variables. For instance, you may only access them in local function and pass to the global function as parameters when needed.

universum/analyzers/uncrustify.py
```diff
@@ -1,200 +1,202 @@
...
+
+    def __call__(self, file: Path, src: List[str], target: List[str]) -> None:
+        file_relative = file.relative_to(Path.cwd())
+        out_file_name: str = str(file_relative).replace('/', '_') + '.html'
+        with open(self.target_folder.joinpath(out_file_name), 'w') as out_file:
+            out_file.write(self.differ.make_file(src, target, context=False))
 
+
+def _uncrustify_output_parser(files: List[Tuple[Path, Path]],
+                              write_diff_file: Optional[Callable[[Path, List[str], List[str]], None]]
```
i-keliukh, line 111: Is "Optional" needed here?

i-keliukh (dismissed):

k-dovgan (approved):

PR #603, opened 2021-05-13
refactor(docker): switch to multistage Dockerfile

k-dovgan:
All other stderr messages cannot be suppressed,
and are both valid and expected.


i-keliukh:
> All other stderr messages cannot be suppressed,
> and are both valid and expected.

What are they?
Can we document them as expected, for example, in the Dockerfile itself?

k-dovgan:
1. `debconf: delaying package configuration, since apt-utils is not installed`
    cannot be removed, as even installing the `apt-utils` itself raises this warning _before_ installation (and is not needed for anything)
2. `WARNING: Running pip as root will break packages and permissions. You should install packages reliably by using venv: https://pip.pypa.io/warnings/venv`
    unlike other pip warnings cannot be suppressed and _in general_ is very true: in most cases installing pip packages as root is considered a bad practise;
   however, in this particular case it still is preferred, as a) root privilege in docker is not that dangerous and b) installing `p4python` can be too long and unstable to perform it in every test repeatedly as a user
3. `2021-05-13 12:02:48 URL:https://bootstrap.pypa.io/get-pip.py [1937346/1937346] -> "get-pip.py" [1]`
    this is not even a warning, it's just printing logs to stderr is expected and documented behaviour for wget;
    can be fully suppressed by `--quiet` or by redirecting stderr to stdout for this command, if needed?

As for the documenting in Dockerfile, I simply do not like the idea of duplicating the same comments to all four files. Maybe you've got other ideas of where to document this?

tests/docker/universum_test_env/Dockerfile
```diff
@@ -1,23 +1,39 @@
...
-RUN wget --no-check-certificate -O get-pip.py 'https://bootstrap.pypa.io/get-pip.py'
+RUN add-apt-repository ppa:deadsnakes/ppa && apt-get update
+RUN apt-get install -y ${PYTHON}-dev ${PYTHON}-distutils gnupg2 libssl-dev build-essential
+
+# Please note: wget is writing logs to stderr, these logs are not any kind of warning
+RUN wget --no-verbose --no-check-certificate -O get-pip.py 'https://bootstrap.pypa.io/get-pip.py'
 RUN ${PYTHON} get-pip.py
 
+FROM universum_test_env_no_vcs AS universum_test_env_no_p4
+ARG PYTHON=python3.7
```
i-keliukh, line 20: Do you need to repeat this line in every stage?

tests/docker/universum_test_env/Dockerfile
```diff
@@ -1,23 +1,39 @@
...
+RUN wget -q http://package.perforce.com/perforce.pubkey -O - | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=1 apt-key add - && \
     echo "deb http://package.perforce.com/apt/ubuntu bionic release" > /etc/apt/sources.list.d/perforce.list && \
-    apt update
+    apt-get update
 
-RUN apt install -y helix-cli
+RUN apt-get install -y helix-cli
 RUN ${PYTHON} -m pip install p4python
+
+# Please note, that using Pip as a root user is, indeed, a bad practice
```
i-keliukh, line 37: Please put this comment before the first pip invocation.

tests/docker/universum_test_env/Dockerfile
k-dovgan: If I don't, the next stage assumes there's no such argument. Not sure if there's a better way to deal with it

i-keliukh (approved):

PR #609, opened 2021-05-18
refactor(args): remove reading from stdin support

k-dovgan:
This option is actually never used and not recommended for usage


i-keliukh:
As far as I remember there is also a test for stdin.

i-keliukh (approved):

Issue #612, opened 2021-05-21
module universum.analyzers.uncrustify can't find files to check


iakoff:
## Description
module universum.analyzers.uncrustify can't find files to check

## Steps to reproduce
both steps executed from folder /home/ciuser/Precommit/temp/
$ /usr/bin/python3.8 -m universum.analyzers.uncrustify --cfg-file /home/ciuser/Precommit/temp/code_style/_cpp.cfg --filter-regex '.*//.(?:c|cpp)' --result-file '${CODE_REPORT_FILE}' --output-directory uncrustify --files .
stderr: Please provide at least one file for analysis
Module sh got exit code 2

$ /usr/bin/python3.8 -m universum.analyzers.uncrustify --cfg-file /home/ciuser/Precommit/temp/code_style/_cpp.cfg --result-file '${CODE_REPORT_FILE}' --output-directory uncrustify --files src/test.c
Module sh got exit code 1

## Expected result
Expected exit with 0 code and code style report to swarm.

## Actual result
exit with unclear failure and no code style report to swarm 

## Environment
Focal
Python 3.8.5
Universum 0.19.7

o-andrieiev:
After some investigation, the original reporter could no longer reproduce the issue.
The analysis of the log files showed that while code_report=True flah is present the reporting step is omitted completely.
This behaviour was deemed impossible before.
One possible causes for this issue is misuse of configuration API. To investigate further, the config file should be rewritten to leave only the code report Step data.

o-andrieiev:
@iakoff please close as unreproducible or provide additional debug data as requested

PR #613, opened 2021-05-21
feat(analyzer): improve usability

o-andrieiev:


o-andrieiev:
This PR does following:
 - informs user to install uncrustify instead of failing (customer request)
 - allows both absolute and relative path strings
 - clarifies failure case, when files don't match the filter pattern (customer request)

o-andrieiev:
this PR does not fix https://github.com/Samsung/Universum/issues/612, but it helps with narrowing down the root cause

universum/analyzers/uncrustify.py
```diff
@@ -1,6 +1,7 @@
 import argparse
 import difflib
 import sys
+from shutil import which
```
k-dovgan, line 4: I think `import shutil` & `shutil.which()` will be easier to read (as it will be more obvious that 'which' is not a locally declared function by reading the line).
The same goes with `os.environ`.
This approach is generally used in most Universum modules

universum/analyzers/uncrustify.py
```diff
@@ -29,15 +33,17 @@ def main() -> int:
     files: List[Tuple[Path, Path]] = []
     try:
         src_files: List[Path] = _get_src_files(settings.file_list, settings.file_list_config)
+        if not src_files:
+            raise EnvironmentError("Please provide at least one file for analysis")
```
k-dovgan, line 37: Why not use Universum-specific exceptions, such as 'CiException' or 'CriticalCiException'?

k-dovgan (approved):

i-keliukh (approved):

PR #615, opened 2021-05-26
fix(docker): bring back noninteractive environment

k-dovgan:
It actually is required only by Python3.9 installation, but nevertheless


k-dovgan:
Only reproduced when rebuilding `universum_test_env` images from scratch for Python3.9; for any other Python version, or if using `make images` instead of `make rebuild`, most likely won't happen.

i-keliukh (approved):

PR #616, opened 2021-05-27
fix(p4): print warning instead of block failing

k-dovgan:
User request due to TeamCity failing build if any blocks fail


k-dovgan:
This is a hotfix of a hotfix; actual solution is expected to be implemented in scope of #617

i-keliukh (approved):

PR #623, opened 2021-06-11
fix(p4): remove empty CL's on force clean

k-dovgan:
Seems that 'client unknown' is enough to stop the whole client cleaning block; while every revert operation should be handled separately, as they all can fail independently.

### DIFF
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -1,6 +1,5 @@
 # pylint: disable = redefined-outer-name
 
-from pathlib import Path
 import pytest
 import P4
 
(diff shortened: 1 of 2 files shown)


tests/test_regression.py
```diff
@@ -134,26 +158,19 @@ def test_p4_api_failed_opened(perforce_environment, mock_opened):
     assert "Getting file diff failed due to Perforce server internal error" in log
 
 
-def test_which_universum_is_tested(docker_main, pytestconfig):
-    config = """
+def test_p4_clean_empty_cl(perforce_environment, stdout_checker):
```
i-keliukh, line 161: Please document the test idea and workflow.
I couldn't figure it out fully from reading the code.

i-keliukh (approved):

PR #625, opened 2021-06-11
feat(analyzer): scan-build report parser

o-andrieiev:
scan-build executes during build process and produces html reports - these reports need to be parsed for code_report to process them


tests/test_code_report.py
```diff
@@ -57,6 +67,8 @@ def finalize(self) -> str:
 
 
 @pytest.mark.parametrize('analyzers, extra_args, tested_content, expected_success', [
+    [['scan_build_report'], [], None, True],
```
o-andrieiev, line 70: in case of scan-build, presence of .html report file already indicates an issue

universum/analyzers/scan_build_report.py
```diff
@@ -0,0 +1,52 @@
...
+
+def scan_build_report_output_parser(file_list: List[str]) -> List[utils.ReportData]:
+    result: List[utils.ReportData] = []
+    for report_file in file_list:
+        with open(report_file) as f:
+            parser = etree.HTMLParser()
+            tree: etree.ElementTree = etree.parse(f, parser)
+            found = any(x.text for x in tree.iter() if x.text == 'Bug Summary')
+            if not found:
+                continue
```
o-andrieiev, line 33: this is to filter out non-report html files - these files don't indicate a problem

universum/analyzers/utils.py
```diff
@@ -4,23 +4,24 @@
...
+    if cmd:
+        result = subprocess.run(cmd, universal_newlines=True,  # pylint: disable=subprocess-run-check
+                                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        if result.stderr and not result.stdout:
+            sys.stderr.write(result.stderr)
+            return result.returncode
 
     try:
-        parsed_issues = parse_function(result.stdout)
+        parsed_issues = parse_function(result.stdout if cmd else '')
```
k-dovgan, line 24: Is there a difference between passing '' and None?

k-dovgan (approved):

universum/analyzers/utils.py
```diff
@@ -4,23 +4,27 @@
...
+        if result.stderr and not result.stdout:
+            sys.stderr.write(result.stderr)
+            return result.returncode
+        data_to_parse = result.stdout
+    else:
+        data_to_parse = ''  # nothing to parse from stdout
 
     try:
-        parsed_issues = parse_function(result.stdout)
+        parsed_issues = parse_function(data_to_parse)
```
i-keliukh, line 27: If I understand the resulting logic of this function correctly it is like this:
1. normally run the specified command and call the function to parse
2. but in some cases instead don't run the command and pass the fake data to parse function.
The only reason I could imagine for this approach is that there is a possibility to reuse the single parsing function for both commandful and commandless scenarios.
Otherwise, it looks more reasonable to just pass the required data (parsed_issues) as an argument and wrap the "subprocess" part into another function to be used by clients. This would also remove the need for the callback.

universum/analyzers/utils.py
```diff
@@ -4,37 +4,88 @@
...
 
-    output_report_to_file(parsed_issues, out_file)
-    return 1 if parsed_issues else 0
+class AnalyzerException(CiException):
+    def __init__(self, code: int = 2, message: Optional[str] = None):
+        self.code: int = code
+        self.message: Optional[str] = message
+
+
+def analyzer(parser: argparse.ArgumentParser):
```
i-keliukh, line 21: If I understand the idea of analyzer and sys_exit decorators correctly, they define the interface between analyzers and CodeReportCollector. Please document this idea (the purpose) in the code.

universum/analyzers/utils.py
```diff
@@ -4,36 +4,91 @@
...
+            issues: List[ReportData] = func(settings)
+            report_to_file(issues, settings.result_file)
+            return issues
+
+        return wrapper
+
+    return internal
+
+
+def sys_exit(func: Callable[[], Any]) -> Callable[[], None]:
```
i-keliukh, line 35: From reading the code it is not clear why 'sys_exit' decorator is separate from 'analyzer' decorator.
sys_exit is only applied to functions that also have analyzer applied.
I suggest to merge them into one.

o-andrieiev: sys_exit is a generic wrapper not limited to analyzers - it has no reference to analyzer logic and can be used for any function/module

universum/analyzers/mypy.py
```diff
@@ -1,31 +1,29 @@
...
     parser = argparse.ArgumentParser(description="Mypy analyzer")
     utils.add_files_argument(parser)
     parser.add_argument("--config-file", dest="config_file", type=str, help="Specify a configuration file.")
     utils.add_python_version_argument(parser)
-    utils.add_result_file_argument(parser)
     return parser
 
 
+@utils.sys_exit
+@utils.analyzer(mypy_argument_parser())
```
i-keliukh, line 17: Please consider changing mypy_argument_parser and friends from functions into global variables.
All of them are only called once.

universum/analyzers/utils.py
i-keliukh: That is ok.
But I don't understand, why 'analyzer' cannot use it internally without the need for clients to apply it explicitly.

universum/analyzers/utils.py
o-andrieiev: yes, this function serves to codify the interface between analyzer and reporter

universum/analyzers/mypy.py
o-andrieiev: I fail to see why that would be a good reason to change functions into globals

universum/analyzers/utils.py
o-andrieiev: it can, but these functions have different purposes logically, and can exist independently.
I see no good reason to add an interdependency here.
E.g. sys_exit is needed only to call the analyzer directly via cmd line. If we add other options to call analyzers, there'd be no need for sys_exit there, but 'analyzer' function would still be needed to preserve the interface.

universum/analyzers/utils.py
o-andrieiev: the code in question has been reworked

universum/analyzers/utils.py
i-keliukh: I would say that we could remove the 'sys_exit' functionality from the 'analyzer' decorator as soon as it is not needed anymore. This is because with the current option (analyzer are scripts) the 'sys_exit' functionality appears to be another part of the interface.
Anyway, it is not strictly required to merge it.

o-andrieiev: another reason why I prefer to keep these decorators separate is that this makes debugging easier - by removing sys_exit from the analyzer code I can easily see the root cause for generated exceptions, even during pytest run

i-keliukh: Please mention this in the comment in the source code

i-keliukh (dismissed):

i-keliukh (approved):

PR #629, opened 2021-06-16
feat(analyzer): add SARIF report

o-andrieiev:


universum/analyzers/sarif_report.py
```diff
@@ -0,0 +1,50 @@
...
+                        # If the region property is absent, the physicalLocation object refers to the entire artifact.
+                        # TODO: cover this case as comment to the file as a whole
+                        line = location_data.get('region').get('startLine')
+                        result.append(utils.ReportData(
+                            symbol="Reported issue",
+                            message=f"{who} : {what}",
+                            path=path,
+                            line=int(line)
+                        ))
+    return result
```
k-dovgan, line 46: I'd add some handling of unsuccessful 'get' calls for user to get more clear understanding of what exactly has gone wrong. This might happen, for example (but not only), if analyzer itself updates and changes the output format

o-andrieiev: the only unchecked 'get' calls are in cases where the property in question SHALL be present in a conformant SARIF file.

i-keliukh:
Did you consider the https://pypi.org/project/sarif-om/ ?

o-andrieiev:
> Did you consider the https://pypi.org/project/sarif-om/ ?

yes, but I don't see how it helps us - we need a very limited set of data put into our own data structures.
SARIF-OM is not even typed - so there is not a lot of difference from just using parsed json.

i-keliukh (approved):

PR #633, opened 2021-06-22
fix(p4): exception message returned by sh.Command in Python3 is byte string, not str

k-dovgan:


universum/modules/vcs/perforce_vcs.py
```diff
@@ -473,10 +473,11 @@ def check_diff_for_depot(self, depot: str) -> str:
             result: str = utils.trim_and_convert_to_unicode(diff_result.stdout)
         except sh.ErrorReturnCode as e:
             for line in e.stderr.splitlines():
-                if not (line.startswith("Librarian checkout")
-                        or line.startswith("Error opening librarian file")
-                        or line.startswith("Transfer of librarian file")
-                        or line.endswith(".gz: No such file or directory")):
+                text = str(line)
```
i-keliukh, line 476: Is it correct way to convert bytes to string?

tests/test_regression.py
```diff
@@ -177,3 +178,24 @@ def test_p4_clean_empty_cl(perforce_environment, stdout_checker):
...
+def test_p4_diff_exception_handling(perforce_environment, mock_diff, stdout_checker):
+    config = """
+from universum.configuration_support import Step, Configuration
+
+configs = Configuration([Step(name="Step", command=["ls"])])
+"""
+    settings = shelve_config(config, perforce_environment)
+    assert __main__.run(settings)
+    stdout_checker.assert_has_calls_with_param("This is error text")
+    # Without the fix all error messages to stderr instead of stdout
```
i-keliukh, line 201: Looks like the sentence is incomplete - the predicate is missing

tests/test_regression.py
```diff
@@ -177,3 +178,24 @@ def test_p4_clean_empty_cl(perforce_environment, stdout_checker):
...
+@pytest.fixture()
+def mock_diff(monkeypatch):
+    def mocking_function(*args, **kwargs):
+        raise sh.ErrorReturnCode(stderr=b"This is error text", stdout=b"This is text",
+                                 full_cmd="any shell call with any params")
+
+    monkeypatch.setattr(sh, 'Command', mocking_function, raising=False)
+
+
+def test_p4_diff_exception_handling(perforce_environment, mock_diff, stdout_checker):
```
i-keliukh, line 192: Is it possible to trigger the bug without the mock?
The mock looks artificial, because you can return virtually anything, not just bytes or string.

k-dovgan: If there is such a way, I couldn't think of it.
To trigger the bug, we need to fail the call of `p4 diff` by `sh`; and this command only fails in very exceptional situations - like, connection interrupted.
And stopping the whole p4 server for this purpose will not only fail all the following Universum execution so much, that tracking this little fail won't be so easy; more importantly, the p4 server shutting up should be timed very carefully to let all the previous operations succeed before this call. I can't think of any better ways of doing this

i-keliukh (approved):

PR #642, opened 2021-07-21
fix(launcher): do not expect execution of Steps without command

k-dovgan:


universum/modules/structure_handler.py
```diff
@@ -138,7 +138,9 @@ def execute_one_step(self, configuration: Step,
         process = step_executor(configuration)
 
         background = configuration.background
-        process.start(is_background=background)
+        is_running = process.start(is_background=background)
```
i-keliukh, line 141: This way the encapsulation of the RunningStep class is broken.
It should be able to handle command absence on its own.

universum/modules/launcher.py
```diff
@@ -184,6 +184,7 @@ def __init__(self, item: configuration_support.Step,
         self.process: sh.RunningCommand
         self._is_background: bool = False
         self._postponed_out: List[Tuple[Callable[[str], None], str]] = []
+        self._needs_finalization = True
```
i-keliukh, line 187: Please add type declaration

universum/modules/launcher.py
```diff
@@ -253,33 +253,38 @@ def add_tag(self, tag: str) -> None:
             self.out.log("Tag '" + tag + "' added to build.")
 
     def finalize(self) -> None:
-        try:
-            text = ""
+        if self._needs_finalization:
```
i-keliukh, line 256: Please invert the condition and add return statement instead of else block

universum/modules/structure_handler.py
k-dovgan: Though just in case I need to point out that without breaking the encapsulation the background steps without command are still handled as actual background steps.

universum/modules/launcher.py
```diff
@@ -202,10 +203,11 @@ def prepare_command(self) -> bool:  # FIXME: refactor
         return True
 
     def start(self, is_background: bool) -> None:
+        self._is_background = is_background
```
i-keliukh, line 206: Can we move _is_background initialization to the constructor?

k-dovgan: We only get the real value from `.start()` function parameters, so if I got the question right — no, I don't think we can

i-keliukh: Read the usage. The distance between constructor call and `.start` call is 3 lines.

i-keliukh (approved):

PR #645, opened 2021-07-28
feat(output): init HTML output feature

miltolstoy:
1. New `--html-log` flag and its processing added
2. Simple tests added
3. New dependencies for tests added to README and setup.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Prerequisites:
    https://docs.docker.com/engine/installation/linux/ubuntu/#install-using-the-repository)
 
    * Also add current user to 'docker' group (use `sudo usermod -a -G docker $USER` and then relogin)
+3. Install Mozilla WebDriver: `sudo apt install firefox-geckodriver`
 
 Further commands:
 ```bash
--- a/tests/test_html_output.py
+++ b/tests/test_html_output.py
@@ -1,0 +1,43 @@
+import os
+import pytest
+
+from selenium import webdriver
+from selenium.webdriver.firefox.options import Options
+
+
+config = f"""
+from universum.configuration_support import Configuration
+
+configs = Configuration([dict(name="Dummy", command=["echo", "dummy"])])
+"""
+
+
+@pytest.fixture()
+def browser():
+    options = Options()
+    options.headless = True
+    browser = webdriver.Firefox(options=options)
+    yield browser
+    browser.close()
+
+
+def test_success(docker_main, browser):
+    docker_main.run(config, additional_parameters="--html-log")
+    log_path = os.path.join(docker_main.artifact_dir, "log.html")
+    assert os.path.exists(log_path)
+
+    browser.get(f"file://{log_path}")
+    body_element = browser.find_element_by_tag_name("body")
+    body_elements = body_element.find_elements_by_xpath(".//*")
+    assert len(body_elements) == 1
+
+    pre_element = body_elements[0]
+    assert pre_element.tag_name == "pre"
+    assert pre_element.text # will be changed to valid HTML in further commits
+
+
+def test_no_html_log_requested(docker_main):
+    docker_main.run(config)
+    log_path = os.path.join(docker_main.artifact_dir, "log.html")
+    assert not os.path.exists(log_path)
+
--- a/tests/test_html_output.py
+++ b/tests/test_html_output.py
@@ -1,3 +1,5 @@
+# pylint: disable = redefined-outer-name
+
 import os
 import pytest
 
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -47,6 +47,7 @@ Prerequisites:
    https://docs.docker.com/engine/installation/linux/ubuntu/#install-using-the-repository)
 
    * Also add current user to 'docker' group (use `sudo usermod -a -G docker $USER` and then relogin)
+3. Install Mozilla WebDriver: `sudo apt install firefox-geckodriver`
```
k-dovgan, line 50: Don't forget to fix the 'universum_log_collapser' section further in Readme file

universum/modules/output/output.py
```diff
@@ -34,40 +39,57 @@ def __init__(self, *args, **kwargs) -> None:
...
 
     def log_stderr(self, line: str) -> None:
         self.driver.log_stderr(line)
+        self.html_driver.log_stderr(line)
 
     def log_shell_output(self, line: str) -> None:
         self.driver.log_shell_output(line)
+        self.html_driver.log_shell_output(line)
+
+    def _create_html_driver(self):
```
k-dovgan, line 89: Why is this function private?

README.md
miltolstoy: I've planned to remove all related to Jenkins plugin in one PR when HTML log functionality will be ready.
Do you want some update or removal right now?

universum/modules/output/output.py
miltolstoy: Because its only purpose is to be called once in `Output` constructor. At least for now.

README.md
k-dovgan: No, separate PR is a good idea, thanks.

k-dovgan (approved):

universum/modules/output/html_output.py
```diff
@@ -0,0 +1,91 @@
...
+    def log_external_command(self, command):
+        self.log(command) # stub
+
+    def log_shell_output(self, line):
+        self.log(line) # stub
+
+    @staticmethod
+    def _is_log_start(line):
+        log_start_pattern = r"^Universum \d+\.\d+\.\d+ started execution$"
+        return re.match(log_start_pattern, line)
```
i-keliukh, line 61: In my opinion this machinery can be replaced by explicit call from the main universum.

i-keliukh (dismissed):

k-dovgan (dismissed):

universum/__main__.py
```diff
@@ -52,7 +52,7 @@ def run(settings: ModuleNamespace) -> int:
     if error_state_module.is_in_error_state():
         raise IncorrectParameterError(("\n\n"+"-"*80 + "\n").join(error_state_module.get_errors()))
 
-    main_module.out.log("{} {} started execution".format(__title__, __version__))
+    main_module.out.log_execution_start(__title__, __version__)
```
i-keliukh, line 55: This call happens before artifacts cleaning.
And artifacts cleaning deletes output html file later.

i-keliukh (changes requested):
Please fix the issue with artifacts cleaning

universum/__main__.py
i-keliukh: Actually, initial artifacts cleaning happens only in nonci mode or if "--clean-build" command line parameter is passed.

tests/test_html_output.py
```diff
@@ -25,7 +25,27 @@ def browser():
...
     docker_main.run(config, additional_parameters="--html-log")
+    check_html_log(docker_main.artifact_dir, browser)
+
+
+def test_success_clean_build(docker_main, browser):
+    docker_main.run(config, additional_parameters="--html-log --clean-build")
+    check_html_log(docker_main.artifact_dir, browser)
+
+
+def test_success_nonci(docker_nonci, browser):
```
i-keliukh, line 36: You can combine this test and 'test_success' by using docker_main_and_nonci fixture

miltolstoy: I don't want to. Different "main" code is executed, so both of these cases should be checked.
And I prefer to follow "one test - one assert" rule (not literally =) ).

tests/test_html_output.py
```diff
@@ -0,0 +1,58 @@
...
+    docker_nonci.run(config, additional_parameters="--html-log")
+    check_html_log(docker_nonci.artifact_dir, browser)
+
+
+def test_no_html_log_requested(docker_main):
```
i-keliukh, line 41: This test can also use common main and nonci fixture

i-keliukh (approved):

PR #649, opened 2021-08-20
feat(output): implement text formatting as in terminal-based output

miltolstoy:


i-keliukh (approved):

universum/modules/output/html_output.py
```diff
@@ -15,68 +15,57 @@ def __init__(self, *args, **kwargs):
         self._filename = None
         self.artifact_dir_ready = False
         self._log_buffer = list()
+        self._block_level = 0
```
k-dovgan, line 18: Are we sure this does not duplicate block leveling, implemented in `structure_handler.py`?

universum/modules/output/html_output.py
```diff
@@ -15,68 +15,57 @@ def __init__(self, *args, **kwargs):
...
     def close_block(self, num_str, name, status):
-        self.log(name) # stub
+        self._block_level -= 1
+        indent = "  " * self._block_level
+        self._log_line(f"{indent} \u2514 [{status}]")
+        self._log_line("")
 
     def report_error(self, description):
-        self.log(description) # stub
+        pass
```
k-dovgan, line 34: Is this behaviour going to be changed in further reviews?

universum/modules/output/html_output.py
```diff
@@ -15,68 +15,57 @@ def __init__(self, *args, **kwargs):
...
-        self.log(line) # stub
+        self._log_line(f"Error: {line}")
 
     def log_stderr(self, line):
-        self.log(line) # stub
+        self._log_line(f"stderr: {line}")
 
     def log(self, line):
-        self._log_line(line) # stub
+        self._log_line(f"==> {line}")
```
k-dovgan, line 52: Isn't this behaviour replicating the `TerminalBasedOutput`? Shouldn't it be integrated in some way?

universum/modules/output/html_output.py
miltolstoy: No, this behavior is copied from `TerminalBasedOutput`.

universum/modules/output/html_output.py
miltolstoy: Yes, common logic will be extracted from `TerminalBasedOutput` and `HtmlOutput` and moved to some parent class. But it will be done after HTML output implementation to avoid double-work.

universum/modules/output/html_output.py
miltolstoy: I'm not sure, I did this similar to `TerminalBasedOutput`. I will check later and reply.

k-dovgan (approved):
Still expect some of these to be refactored in further development.

PR #653, opened 2021-09-01
feat(output): add sections collapsing in HTML output

miltolstoy:
Wrapping HTML example:
```
<input type="checkbox" id="1." class="hide">
<label for="1.">
<span class="sectionLbl">1. Section name</span>
</label>
<div>Section body</div>
```

Expected behavior:
* all steps are collapsed at HTML opening
* click on collapsed section name leads to:
    - section body is shown, inner sections are collapsed
    - section marker changed from `[+]` to `[-]`
* click on not collapsed section name leads to opposite change

### DIFF
--- a/universum/modules/output/html_output.py
+++ b/universum/modules/output/html_output.py
@@ -73,7 +73,8 @@ class HtmlOutput(BaseOutput):
         self._log_line(html_footer)
 
     def _log_line(self, line, with_line_separator=True):
-        line = f"{line}{os.linesep}" if with_line_separator else line
+        if with_line_separator and not line.endswith(os.linesep):
+            line += os.linesep
         if not self._filename:
             raise RuntimeError("Artifact directory was not set")
         if not self.artifact_dir_ready:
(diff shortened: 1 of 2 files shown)


universum/modules/output/html_output.py
```diff
@@ -67,7 +72,8 @@ def log_execution_finish(self, title, version):
         html_footer = "</pre></body></html>"
         self._log_line(html_footer)
 
-    def _log_line(self, line):
+    def _log_line(self, line, with_line_separator=True):
+        line = f"{line}{os.linesep}" if with_line_separator else line
```
i-keliukh, line 76: It looks like the line separator should not be added in case the line already has one.
At least this logic is implemented in the terminal output.
Without it, the html output has the empty line after the "Here is the summarized build result:" string.

miltolstoy: Fixed.

i-keliukh (approved):

PR #654, opened 2021-09-08
refactor(pylint): updated according to new recommendations

k-dovgan:
1. open() is recommended to use with specific encoding
   instead of relying on autodetect
2. list and dict literals are recommended for use instead of
   constructors for optimization reasons
3. 'u""' prefix is no longer required, as strings are unicode
   by default

### DIFF
--- a/universum/modules/error_state.py
+++ b/universum/modules/error_state.py
@@ -41,7 +41,7 @@ class HasErrorState(Module):
             return ""
         if value.startswith('@'):
             try:
-                with open(value.lstrip('@')) as value_file:
+                with open(value.lstrip('@'), encoding="utf-8") as value_file:
                     return value_file.read()
             except FileNotFoundError as e:
                 self.error(f"Error reading argument {setting_name} from file {e.filename}: no such file")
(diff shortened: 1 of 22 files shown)


universum/modules/error_state.py
```diff
@@ -41,7 +41,7 @@ def read_multiline_option(self, setting_name: str) -> str:
             return ""
         if value.startswith('@'):
             try:
-                with open(value.lstrip('@')) as value_file:
+                with open(value.lstrip('@'), encoding="utf-8") as value_file:
```
i-keliukh, line 44: Shouldn't we document this change?

universum/modules/launcher.py
```diff
@@ -358,7 +358,7 @@ def process_project_configs(self) -> configuration_support.Configuration:
         sys.path.append(os.path.join(os.path.dirname(config_path)))
 
         try:
-            with open(config_path) as config_file:
+            with open(config_path, encoding="utf-8") as config_file:
```
i-keliukh, line 361: Is there documented way to specify the encoding for .py files?
If yes, we should consider using it instead of hardcoding.

i-keliukh: Actually, PEP263 still works in Python 3. The only change is the default encoding. 

Please add the requirement for utf8 for configs to documentation.

universum/modules/error_state.py
i-keliukh: What about this change?

i-keliukh (approved):

PR #656, opened 2021-09-14
feat(report): move SARIF to report format

o-andrieiev:


tests/test_code_report.py
```diff
@@ -23,13 +23,20 @@ def __init__(self):
...
         self.text +=\
-            f"configs += Configuration([Step(name='Run {analyzer}', {extra_cfg} code_report=True, command={cmd})])\n"
+            f"configs += Configuration([Step(name='{name}', command={cmd}{step_cfg})])\n"
         return self
 
+    def add_analyzer(self, analyzer: str, arguments: List[str], step_cfg: str = '') -> 'ConfigData':
+        name = f"Run {analyzer}"
+        args = [f", '{arg}'" for arg in arguments]
+        cmd = f"['{python()}', '-m', 'universum.analyzers.{analyzer}'{''.join(args)}]"
+        step_cfg = ', ' + step_cfg if step_cfg else ''
```
k-dovgan, line 36: Is there a reason we're using `cfg` instead of `config` as in other modules like `Launcher`?

o-andrieiev: I have no preference either way, I'll change it to 'config' then. But please note that this is a test case code.

k-dovgan (dismissed):

o-andrieiev:
Retest

i-keliukh (changes requested):
1. sarif_report must be removed from test parameter on line 218, file test_code_report.py
2. please rewrite asserts without findall calls to simplify log analysis. pytest is able to produce assertion failure logs, if the asserted expression is simple enough.

i-keliukh:
One variant of making the assert output readable is to separate the check and the assert and add the assert message, like below:
```python
if not re.findall(log_success, log):
    assert False, f"'{log_success}' is not found in '{log}'"
```

tests/test_code_report.py
```diff
@@ -118,11 +130,40 @@ def finalize(self) -> str:
...
+    'both_tested_issues_in_json',
+])
+def test_code_report_direct_log(runner_with_analyzers, tested_contents, expected_success):
+    config = ConfigData()
+    step_config = "code_report=True"
+    for idx, tested_content in enumerate(tested_contents):
+        prelim_report = "report_file_" + str(idx)
+        full_report = "${CODE_REPORT_FILE}"
+        runner_with_analyzers.local.root_directory.join(prelim_report).write(tested_content)
+        config.add_cmd("Report " + str(idx), f"[\"bash\", \"-c\", \"cat ./{prelim_report} >> {full_report}\"]",
```
i-keliukh, line 159: The paths to files must be wrapped in quotes.

i-keliukh (approved):

PR #660, opened 2021-09-21
feat(test): add a variable to cache docker containers outside PyCharm

k-dovgan:
.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,10 @@ nox
 This will launch the testing scenario, described in `noxfile.py`. This scenario includes rebuilding docker images
 for every supported Python version and running all the tests for corresponding Python.
 
+Also, setting up "REUSE_DOCKER_CONTAINERS" environment variable (or running tests in PyCharm) will let tests
+run in docker cache and reuse test containers. But do note that this is recommended for development purposes only,
+as recreating clear containers make testing environment more dependable.
+
 
 ## Project contents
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -110,6 +110,10 @@ nox
 This will launch the testing scenario, described in `noxfile.py`. This scenario includes rebuilding docker images
 for every supported Python version and running all the tests for corresponding Python.
 
+Also, setting up "REUSE_DOCKER_CONTAINERS" environment variable (or running tests in PyCharm) will let tests
+run in docker cache and reuse test containers. But do note that this is recommended for development purposes only,
```
i-keliukh, line 114: Tests are not run in docker cache.
The correct way to say this would be: "Tests reuse already created and initialized containers, which speeds up the testing process".

README.md
```diff
@@ -110,6 +110,10 @@ nox
...
+as recreating clear containers make testing environment more dependable.
```
i-keliukh, line 115: What is meant by "more dependable"?

README.md
k-dovgan: It was "tests run in docker" "cache and reuse container". But ok, I'll rephrase that.

README.md
i-keliukh: Just say it directly: "Without recreating containers, the remnants of previous test runs may affect the current test run."

i-keliukh (approved):

PR #663, opened 2021-09-27
feat(p4): resolve multiple unshelved CLs

### LINKED ISSUE #648: Universum should merge pending changelists instead of forcibly rewriting one another
Universum uses "-f" option together with "p4 unshelve" sub-command:

- https://github.com/Samsung/Universum/blob/01f4433f0a6c66cc429f4db171c4ea9675a13746/universum/modules/vcs/perforce_vcs.py#L467

This means that if we have multiple shelved changelists that modify the same file, only the last unshelved CL will be taken into account.
It means that other CLs will be silently ignored.

For example, instead of:
```
p4 unshelve -s CL_1 -f CL_TMP
p4 unshelve -s CL_2 -f CL_TMP
```

it would be better to use:
```
p4 unshelve -s CL_1 -c CL_TMP
p4 unshelve -s CL_2 -c CL_TMP

<analyzing P4 output>
<decide to return ERROR or resolving >

p4 resolve -am
```

IMHO, the current quiet behavior can be confusing to the user.

k-dovgan:
Resolves #648

### DIFF
--- a/universum/modules/vcs/perforce_vcs.py
+++ b/universum/modules/vcs/perforce_vcs.py
@@ -459,6 +459,26 @@ class PerforceMainVcs(PerforceWithMappings, base_vcs.BaseDownloadVcs):
             raise
         return result
 
+    def p4resolve(self):
+        info = {}
+
+        try:
+            info = self.p4.run_resolve("-am")
+            if "resolve skipped" in str(info):
+                error_message = "Problem during merge while resolving shelved CLs!"
+                self.out.log(error_message)
+                self.out.log(str(info))
+                raise Exception(error_message)
+
+        except P4Exception as e:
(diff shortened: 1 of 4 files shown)


universum/modules/vcs/perforce_vcs.py
```diff
@@ -459,6 +459,13 @@ def p4unshelve(self, *args, **kwargs):
             raise
         return result
 
+    @catch_p4exception(ignore_if="No file(s) to resolve")
+    def p4resolve(self):
+        result = str(self.p4.run_resolve("-am"))
```
i-keliukh, line 464: In my opinion this operation and its outcome must be reported to both logs and repo status.

PR #682, opened 2021-11-02
feat(p4): add login to 'p4.connect()'

k-dovgan:


i-keliukh:
Please add proper description.
Right now it is not clear what we are doing and why we are doing it.

k-dovgan:
To reproduce the bug fixed by the latest fix, execute `universum run`. When not executing from config file (e.g. via simple `make test` or `pytest -k`), all tests passed. Not exactly sure right now what causes the difference in behaviour.

tests/perforce_utils.py
```diff
@@ -159,12 +159,12 @@ def perforce_connection(request, docker_perforce):
...
     p4.connect()
     p4.run_login()
-    yield p4
+    yield utils.Params(p4=p4, password=docker_perforce.password)
     p4.disconnect()
 
 
 class PerforceWorkspace(utils.BaseVcsClient):
-    def __init__(self, connection: P4, directory: py.path.local):
+    def __init__(self, connection: P4, password: str, directory: py.path.local):
```
i-keliukh, line 167: Please remove the password from the PerforceWorkspace.
It should only provide our "connection" class with password, not store the password on its own.

tests/perforce_utils.py
```diff
@@ -280,7 +281,7 @@ def cleanup(self) -> None:
 
 @pytest.fixture()
 def perforce_workspace(request, perforce_connection, tmpdir):
-    workspace = PerforceWorkspace(perforce_connection, tmpdir)
+    workspace = PerforceWorkspace(perforce_connection.p4, perforce_connection.password, tmpdir)
```
i-keliukh, line 284: Please change the code to pass (and store) the full perforce_connection object, instead of just P4 instance.

i-keliukh (approved):

i-keliukh (approved):

PR #683, opened 2021-11-03
feat(output): log_stderr() HTML coloring

miltolstoy:
1. Coloring stderr output to orange color, as it was done in `TerminalBasedOutput`
2. launcher.py updated to write stderr logs to HTML also, if write to file was selected

### DIFF
--- a/universum/modules/output/html_output.py
+++ b/universum/modules/output/html_output.py
@@ -55,7 +55,7 @@ class HtmlOutput(BaseOutput):
         self._log_line(f'<span class="exceptionTag">Error:</span> {line}')
 
     def log_stderr(self, line):
-        self._log_line(f"stderr: {line}")
+        self._log_line(f'<span class="stderrTag">stderr:</span> {line}')
 
     def log(self, line):
         self._log_line(f"==> {line}")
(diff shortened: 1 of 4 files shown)


k-dovgan (approved):

universum/modules/output/html_output.py
```diff
@@ -141,6 +141,9 @@ def _build_html_head():
             .exceptionTag {
                 color: darkred;
             }
+            .stderrTag {
+                color: orange;
```
i-keliukh, line 145: In my opinion this color is too bright for white background.
It is ok on a black background.

k-dovgan (changes requested):
Actually, I don't like the fact, that we are duplicating only stderr from file into HTML log. I thought, we are duplicating all the info.

Only duplicating stderr for commands that actively use both stderr and stdout streams for user interaction (such as e.g. Pylint) will lead to very confusing outcome.

universum/modules/output/html_output.py
miltolstoy: As discussed in chat, it looks +/- okay. Let's leave it as is for now.

i-keliukh (approved):

k-dovgan (approved):

PR #687, opened 2021-11-05
refactor(tests): remove class 'Params', use strict typing instead

k-dovgan:


tests/perforce_utils.py
```diff
@@ -108,6 +108,14 @@ def create_new_p4_container(request, client, name):
         request.raiseerror("Timeout on waiting perforce server to start.")
 
 
+class PerfoceDockerContainer:
```
i-keliukh, line 111: You should be able to use this class as annotation in the usage of the fixture, like follows:
```
def perforce_connection(request, docker_perforce: PerfoceDockerContainer):
```

tests/perforce_utils.py
```diff
@@ -153,6 +160,12 @@ def docker_perforce(request):
                     pass
 
 
+class PerforceConnection:
```
i-keliukh, line 163: The "Any" is still used in the __init__ method of the PerforceWorkspace class. This should be fixed.

tests/deployment_utils.py
```diff
@@ -152,6 +152,13 @@ def clean_execution_environment(request):
             runner.exit()
 
 
+class LocalSources(utils.BaseVcsClient):
```
i-keliukh, line 155: What is the idea of introducing the class and not using it in type annotations?

tests/deployment_utils.py
```diff
@@ -178,24 +180,27 @@ def local_sources(tmpdir):
...
+        self.git: GitClient = git_client
+        self.local: LocalSources = local_sources
+        self.nonci: bool = nonci
 
         # Need to be initialized in 'set_environment'
-        self.environment = None
-        self.working_dir: str = None
-        self.project_root = None
-        self.artifact_dir = None
+        self.environment: ExecutionEnvironment = None  # type: ignore
```
i-keliukh, line 193: Why "type: ignore" instead of Optional?

k-dovgan: Because otherwise Mypy generates lots of warnings later that are basically 'your object might be None and you cannot treat it like ExecutionEnvironment/py.path.local/etc.'
And these warnings would be correct if we didn't manually make sure to set all this objects to `not None` before calling any function that may refer to them.

i-keliukh:
1. There are multiple usages of perforce_workspace fixture without annotation.
2. The parameters of runner_without_environment fixture are not annotated.

k-dovgan:
>     1. There are multiple usages of perforce_workspace fixture without annotation.
> 
>     2. The parameters of runner_without_environment fixture are not annotated.

'runner_without_environment' is the only place I found where 'perforce_workspace' was not annotated. Could you please point out the other ones?

i-keliukh:
```
tests/test_p4_exception_handling.py:12:def perforce_environment(perforce_workspace, tmpdir):
tests/test_p4_revert_unshelved.py:12:    def __init__(self, perforce_workspace):
tests/test_p4_revert_unshelved.py:13:        self.perforce_workspace = perforce_workspace
tests/test_p4_revert_unshelved.py:51:def diff_parameters(perforce_workspace):
tests/test_p4_submit.py:9:def p4_submit_environment(perforce_workspace, tmpdir):
tests/test_poll.py:14:def test_p4_success_command_line_no_changes(stdout_checker, perforce_workspace, tmpdir):
tests/test_poll.py:40:def test_p4_error_command_line_wrong_port(stdout_checker, perforce_workspace, tmpdir):
tests/test_poll.py:67:def poll_environment(request, perforce_workspace, git_client, tmpdir):
tests/test_regression.py:84:def perforce_environment(perforce_workspace, tmpdir):
tests/test_submit.py:22:def p4_submit_environment(perforce_workspace, tmpdir):
tests/test_submit.py:120:def submit_environment(request, perforce_workspace, git_client, tmpdir):
tests/test_unicode.py:16:def test_unicode(vcs, test_type, perforce_workspace, git_client, unicode_dir):
tests/test_git_poll.py:9:def git_poll_environment(git_client, tmpdir):
tests/test_report.py:33:def report_environment(git_client, tmpdir):
tests/test_submit.py:120:def submit_environment(request, perforce_workspace, git_client, tmpdir):
tests/test_integration.py:317:def test_abort(local_sources, tmpdir, terminate_type):
```

i-keliukh (approved):

PR #689, opened 2021-11-09
feat(output): expand failed HTML sections

miltolstoy:
1. On page open all failed sections and their parents will be automatically expanded.
2. CSS and JS sources moved to separate files to clean up the `HtmlOutput` class
    2.1. CSS sources were moved without change
    2.2. JS sources were moved from the old plugin and updated to current needs
Screenshot 2021-11-09 at 10-49-08 Screenshot

### DIFF
--- a/universum/modules/output/html_output.py
+++ b/universum/modules/output/html_output.py
@@ -17,7 +17,7 @@ class HtmlOutput(BaseOutput):
         self.artifact_dir_ready = False
         self._log_buffer = []
         self._block_level = 0
-        self.current_dir = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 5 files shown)


universum/modules/output/html_output.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, *args, **kwargs):
         self.artifact_dir_ready = False
         self._log_buffer = []
         self._block_level = 0
+        self.current_dir = os.path.dirname(os.path.realpath(__file__))
```
i-keliukh, line 20: In my opinion, "current_dir" does not entirely describe the meaning of this variable.
Probably, "module_dir" would be better name.

i-keliukh (approved):

PR #691, opened 2021-11-15
feat(output): parametrize HTML log name

miltolstoy:
**Before**:
    1. `python3 -m universum` - HTML log disabled
    2. `python3 -m universum --html-log` - HTML log saved to file with the default name 
**Now**:
    + one more option 
    3. `python3 -m universum --html-log custom` - HTML log saved to file with the custom name

### DIFF
--- a/tests/test_html_output.py
+++ b/tests/test_html_output.py
@@ -61,9 +61,9 @@ def browser():
 
 
 def test_cli_log_custom_name(tmpdir):
-    log_name = "custom_name"
-    artifact_dir = check_cli(tmpdir, ["-hl", log_name])
-    assert os.path.exists(os.path.join(artifact_dir, f"{log_name}.html"))
+    custom_log_name = "custom_name"
+    artifact_dir = check_cli(tmpdir, ["-hl", custom_log_name])
+    assert os.path.exists(os.path.join(artifact_dir, f"{custom_log_name}.html"))
 
 
 def test_cli_log_default_name(tmpdir):
--- a/tests/test_html_output.py
+++ b/tests/test_html_output.py
@@ -68,7 +68,7 @@ def test_cli_log_custom_name(tmpdir):
 
 def test_cli_log_default_name(tmpdir):
     artifact_dir = check_cli(tmpdir, ["-hl"])
-    assert os.path.exists(os.path.join(artifact_dir, f"universum_log.html"))
+    assert os.path.exists(os.path.join(artifact_dir, "universum_log.html"))
 
 
 def test_cli_no_log_requested(tmpdir):
--- a/tests/test_html_output.py
+++ b/tests/test_html_output.py
@@ -35,11 +35,13 @@ configs = \
     failed_critical_step + success_step
 """
 
+log_name = "universum_log"
+
 
 def create_environment(test_type, tmpdir):
     env = utils.LocalTestEnvironment(tmpdir, test_type)
     env.configs_file.write(config)
-    env.settings.Output.html_log = True
+    env.settings.Output.html_log_name = log_name
     return env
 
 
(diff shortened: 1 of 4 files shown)


k-dovgan:
1. I don't like the option name: 'html log name' does not make it clear that this IS the option that turns logging on.
2. Is there a possibility to have a default log name in this case?

k-dovgan:
Are you intending to add checks of this option to 'test_argument_check.py'?

universum/modules/output/html_output.py
```diff
@@ -11,16 +11,22 @@
...
         super().__init__(*args, **kwargs)
-        self._filename = None
+        self._log_name = self.default_log_name
+        self._log_path = None
         self.artifact_dir_ready = False
         self._log_buffer = []
         self._block_level = 0
         self.module_dir = os.path.dirname(os.path.realpath(__file__))
 
+    def set_log_name(self, log_name):
```
k-dovgan, line 25: Do we really need these setters? Can we maybe pass the params to init?..

miltolstoy: `set_log_name()` - removed
`set_artifact_dir()` - yes. `HtmlOutput` module is created under `Output` class, but only `ArtifactCollector` class is aware of the artifact directory path for the current build.

k-dovgan (dismissed):

universum/modules/output/output.py
```diff
@@ -22,15 +22,24 @@ class Output(Module):
     terminal_driver_factory = Dependency(TerminalBasedOutput)
     html_driver_factory = Dependency(HtmlOutput)
 
+    html_log_default_name = "universum_log"
```
i-keliukh, line 25: I think it would be better to keep the default log file name in the HtmlOutput class.
In case user didn't supply the name, the instance of the HtmlOutput should be created without any parameter.
They it should be able to decide what to do on its side.

universum/modules/output/output.py
```diff
@@ -22,15 +22,24 @@ class Output(Module):
...
+    html_log_disabled_arg_value = "disabled"
```
i-keliukh, line 26: In my opinion, the value for the disabling the log should be chosen more carefully.
With current value user can run the following command that doesn't produce html log:
```
python -m universum ... --html-log disabled
```
As a user I would expect that there are no "special" undocumented file names that I cannot use for my html logs.

tests/test_html_output.py
```diff
@@ -57,6 +60,23 @@ def browser():
     firefox.close()
 
 
+def test_cli_log_custom_name(tmpdir):
+    log_name = "custom_name"
```
i-keliukh, line 64: You are mutating the global variable inside test.
What if the execution order of tests changes?

miltolstoy: There is no `global log_name` line at the start of the test, so the `log_name` global variable will be not changed.
But I've updated the variable name to clarify that it's local.

universum/modules/output/output.py
```diff
@@ -22,15 +22,23 @@ class Output(Module):
     terminal_driver_factory = Dependency(TerminalBasedOutput)
     html_driver_factory = Dependency(HtmlOutput)
 
+    html_log_disabled_arg_value = "__html_log_disabled_arg_value"
```
i-keliukh, line 25: Is it possible to use non-string value here? Like None of False.

tests/test_html_output.py
i-keliukh: Ah, I was confused by the variable name. Thank you for the fix.

universum/modules/artifact_collector.py
```diff
@@ -13,7 +13,7 @@
 from ..lib.utils import make_block
 from ..lib import utils
 from .automation_server import AutomationServerForHostingBuild
-from .output import HasOutput
+from .output import Output, HasOutput
```
i-keliukh, line 16: This change is not needed anymore

i-keliukh (approved):

k-dovgan:
Looks good to me.

PR #692, opened 2021-11-26
fix(pylint): allow forward declarations

miltolstoy:
According to PEP-484, forward declarations have to be strings.
Applied change implicitly turns all annotations to strings.


k-dovgan (approved):

i-keliukh (approved):

k-dovgan:
Looks good to me.

PR #693, opened 2021-11-30
feat(docs): instructions on integrating HTML log with Jenkins

k-dovgan:
https://universum.readthedocs.io/en/add-notes-on-log-usage/jenkins.html


miltolstoy (commented):
LGTM

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
+Integrate collapsible log into Jenkins
+======================================
+
+The `--html-log <args.html#Output>`__ option enables generating an HTML page with collapsible blocks,
```
o-andrieiev, line 4: I don't like 'enables generating'
generates?

doc/args.rst
```diff
@@ -34,5 +34,9 @@ add comments on found issues right to the selected code review system.
         | * -f='test 1:!unit test 1'    - run all steps with 'test 1' substring in their names except those
          containing 'unit test 1'
 
+    --html-log -hl : @after
+        To integrate this log into Jenkins job with ``htmlpublisher`` plugin,
```
o-andrieiev, line 38: it's not clear at all from here why would I want to do that

o-andrieiev: How about
"Generates user-friendly self-contained HTML log. To integrate this log with blabla see blabla"
?

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+very convenient when navigating a long build log of a `Universum` run with many steps.
+
+While TeamCity has its own mechanisms increasing readability, and Universum ran locally by default just stores
```
o-andrieiev, line 7: This is a really long sentence, which is hard to parse.
Split into multiple shorter ideas. E.g.
By default, Jenkins outputs full build log as a plain text without navigation anchors. To simplify interaction and debug, we provide a way to integrate generation of a user-friendly log into a Universum job.
Note: this mechanism is not needed for certain use cases. TeamCity has its own mechanism blabla. Running Universum locally blabla.

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+step execution logs to files to be accessed separately, Jenkins just provides the full build log, formatted as
+a very long text with no real anchors. Generating a user-friendly log and integrating it into a `Universum` job
+simplifies interaction with `Univesrum` in CI process.
+
+One of possible ways to integrate an existing log to Jenkins job is `htmlpublisher
```
o-andrieiev, line 12: Just omit the 'One of possible ways'
To integrate an existing log to Jenkins job use blabla. After installing it, apply blabla.

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+<https://plugins.jenkins.io/htmlpublisher/>`__ plugin. After installing it to your server, apply
+the following changes:
+
+1. Add ``-hl``/``--html-log``/``-hl <name>`` option to the `Universum` command line
+2. Pass default or customized log name to plugin configuration as described in manual
```
o-andrieiev, line 17: Just omit 'default or customized', unless it has to be expressly spelled this way?

o-andrieiev: otherwise "default or customized" just means "any"?

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+   (https://plugins.jenkins.io/htmlpublisher/) via `Post-build Actions` or pipeline
+3. Let Jenkins render the interactive content of log (see https://www.jenkins.io/doc/book/security/user-content/
+   details)
+
+The easiest way to execute the last step proposed above is to `configure Jenkins to publish artifacts on a different
```
o-andrieiev, line 22: Replace "the easiest way to execute the last step proposed above" with "To do this confiugure blabla" and add it as a part of third bullet

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+domain <https://www.jenkins.io/doc/book/security/user-content/#resource-root-url>`__ by changing ``Resource Root URL``
+in `Manage Jenkins » Configure System » Serve resource files from another domain` from something like
+``my.jenkins.io`` to ``res.my.jenkins.io``.
+
+
+Using subdomain with Nginx reverse-proxy
+----------------------------------------
+
+Please pay attention, that server interaction with resource subdomain is not done via ``localhost`` network interface,
```
o-andrieiev, line 31: Replace "Please pay attention, that" with "Note that"

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+but involves full scale redirecting to an external resource. This means that:
```
o-andrieiev, line 32: I am not sure whay "full scale redirecting" is

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+
+1. Both ``my.jenkins.io`` and ``res.my.jenkins.io`` domain names should be resolved correctly
```
o-andrieiev, line 34: "should be" -> "must be" or "shall be", use RFC 2119 terminology in technical docummentation

o-andrieiev: btw, I am not sure what's the intention of this line. Does it mean that user/client has to check and verify that this condition is satisfied?

o-andrieiev: Can we provide the steps to check that?

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+2. All headers should be passed correctly
+
+In case your Jenkins server is launched behind the Nginx reverse-proxy (which is a quite common configuration),
```
o-andrieiev, line 37: Omit "which is a quite common configuration"

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+default behaviour is to replace ``Host`` headers on redirection. To pass them correctly, adjust the configuration
```
o-andrieiev, line 38: what does it mean "default behaviour"? whose behaviour?
Suggested wording:
"In case your Jenkins server is launched behind the blabla, it may replace the host headers on redirection. See X on how to fix this.

doc/html_publisher.rst
```diff
@@ -0,0 +1,44 @@
...
+accordingly (see https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/#passing-request-headers for
+more details).
+
+.. note::
+
+    Also, setting the ``server_name`` to both main domain and a subdomain might be needed
```
o-andrieiev, line 44: Suggested wording:
Also you might need to set server_name to both main domain and a subdomain.

o-andrieiev: Btw, I'm not sure how you can set something to 2 different things, but I might be misreading this. Did you mean "set for"?

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
+Integration with Jenkins
+========================
+
+In `Jenkins` `Universum` is usually launched via single ``{python} -m universum`` command as one long step in a
```
o-andrieiev, line 4: I am not sure about "In Jenkins Universum is launched". Is it run IN Jenkins?

k-dovgan:
This is a test comment, please ignore.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+
+A recommended way to allow Jenkins server to render interactive user content is to `configure Jenkins to publish
+artifacts on a different domain <https://www.jenkins.io/doc/book/security/user-content/#resource-root-url>`__
+by changing ``Resource Root URL`` in `Manage Jenkins » Configure System » Serve resource files from another domain`
+from something like ``my.jenkins.io`` to ``res.my.jenkins.io``.
+
+Note that Jenkins interaction with resource domain, located on the same host is not done via ``localhost``
+network interface. Jenkins treats ``another domain`` as an external URL when redirecting. This means that:
+
+1. Both ``my.jenkins.io`` and ``res.my.jenkins.io`` domain names must be resolved correctly
```
i-keliukh, line 36: On what machine?
Without this context this implies that they must resolve globally.
I guess this is possible, but we cannot verify it.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+2. All headers must be passed correctly
```
i-keliukh, line 37: From where to where?
What "all headers"?

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+
+.. note::
+
+    If main server domain name is not resolved using DNS, ``/etc/hosts`` or any other means, the web-interface
+    will only be accessible via IP, and not the name. As resource domain might be located at the same IP,
+    the resolving of both names is crucial
```
i-keliukh, line 43: Why it is crucial?
The connection between sentence 1 and 2 in this note is not clear.
I think that to make it clear, you can say that the server distinguishes the main domain and the resource domain by the host name.
This is pretty standard way of implementing virtual hosting.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Here are the symptoms of domain names not resolving correctly:
```
i-keliukh, line 45: I propose to move symptoms or issues section to the end of the document.
In my opinion, this doc would be more usable if it contained instructions of what should be done to make domain names resolve correctly.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+
+1. Jenkins warnings when trying to save the updated settings
+2. Client inability to access said pages (timeout error)
+
+A reason for headers not passed correctly can be Nginx configuration. Without specification Nginx replaces
```
i-keliukh, line 50: This is the first time the doc mentions nginx.
However, from the paragraph it looks like it assumes the reader already knows why we use nginx.
I suggest to mention nginx in the intro section on this page.
We can assume that it is present and setup as a reverse proxy, but have to declare it explicitly.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+``Host`` headers on redirection, that will lead to ``404 NOT FOUND`` errors when trying to access artifacts
```
i-keliukh, line 51: nginx doesn't do redirection.
It performs proxying.
BTW, you can explain why 404 is returned if the configuration is incorrect.
This would give the reader more context and may be improve the understanding of this setup.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+from Jenkins. To pass them correctly, adjust the configuration accordingly (see
+https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/#passing-request-headers for more details).
+
+.. note::
+
+    Also you might need to set ``server_name`` to ``my.jenkins.io res.my.jenkins.io`` if they are located
```
i-keliukh, line 57: Technically user can create two separate "servers" in nginx config.
On the other hand, it is not clear how Jenkins can support hosting resource domain on separate host.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+    on the same host
+
+
+Further integration
+-------------------
+
+To integrate the generated log into `Jenkins` job, we suggest `htmlpublisher <https://plugins.jenkins.io/htmlpublisher/>`__
```
i-keliukh, line 64: The plugin is called "HTML Publisher"

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
...
+plugin. After installing it to your server, apply the following changes:
+
+1. Add ``-hl``/``--html-log``/``-hl <name>`` option to the `Universum` command line
+2. Pass the log name to plugin configuration as described in manual (https://plugins.jenkins.io/htmlpublisher/)
+   via `Post-build Actions` or pipeline
+3. Let Jenkins render the interactive content of log (the section above)
```
i-keliukh, line 70: I suggest to move the "plugin" section before "resource url" section.
It is totally possible that reader already has the resource url setup correctly and he doesn't need to read all of this just to find the HTML Publisher section.

doc/jenkins.rst
```diff
@@ -0,0 +1,70 @@
+Integration with Jenkins
```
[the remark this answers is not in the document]
k-dovgan: The title is changed to match the similar page "Integration with TeamCity".
I do not see other info on Jenkins integration right now, but if we come across such, it definitely will be added to this page.
As for the guide, it is a very different part of the documentation, that has nothing to do with _reverse proxy configs_ and such. But the contents of this page will definitely be mentioned in the guide (as well, as there are now mentioned the contents of 'GitHub App' manual).

doc/jenkins.rst
k-dovgan: Well, to be completely honest, it is not that clear for me either.

i-keliukh: Then you can safely assume they are always located on the same host and there is no need to additionally mention it.

doc/jenkins.rst
k-dovgan: I totally agree about instructions, but I'd either remove 'symptoms' (proposed by Oleksandr) completely, or leave them in this paragraph, as they only apply to domain name resolving, and make no sense outside of it

k-dovgan:
This is a test comment, please ignore.

doc/jenkins.rst
```diff
@@ -0,0 +1,136 @@
...
+
+Compare with these:
+
+* When ran locally, all Universum build step logs are stored to separate files instead of printing their
+  output to console
+* When launched by TeamCity, `Unviersum` uses `service messages
+  <https://www.jetbrains.com/help/teamcity/service-messages.html>`__ to increase log readability
+
+To simplify navigating long logs and finding relevant information on build results, we provide a user-friendly
+interactive log with collapsible blocks and other features.
```
i-keliukh, line 16: Here you can mention that the log is usable even outside of Jenkins, but for Jenkins it provides missing functionalty.

doc/jenkins.rst
```diff
@@ -0,0 +1,136 @@
...
+.. _jenkins#plugin:
+
+Add a Jenkins plugin
+--------------------
+
+A `HTML Publisher <https://plugins.jenkins.io/htmlpublisher/>`__ plugin is a very convenient way to add an HTML report
+to a Jenkins job. To use it, you will need to:
+
+1. Install it on server (server restart might be needed to apply changes)
+2. Add ``-hl``/``--html-log``/``-hl <name>`` option to the `Universum` command line as mentioned in previous section
```
i-keliukh, line 58: I think you can remove this item if it is already mentioned in the previous section

doc/jenkins.rst
```diff
@@ -0,0 +1,136 @@
...
+
+Note that Jenkins interaction with resource domain, resolved to the same host IP is not done via ``localhost``
+network interface. The reason for that is Jenkins requiring some interaction with itself via this domain name.
+This means that both ``my.jenkins.io`` and ``res.my.jenkins.io`` domain names must be resolved correctly, either
+globally (via DNS) or locally on both client and server machines (via ``/etc/hosts`` files).
+
+.. note::
+
+    If main server domain name is not resolved using DNS, ``/etc/hosts`` or any other means, the web-interface
+    will only be accessible via IP, and not the name. Because of that, the mentioned above Jenkins interaction
```
i-keliukh, line 87: I propose to remove "mentioned above".
Also you can mention why it will fail (because host name is not passed to the jenkins server).

doc/jenkins.rst
```diff
@@ -0,0 +1,136 @@
...
+As already mentioned above, due to `Jenkins Content-Security-Policy
+<https://www.jenkins.io/doc/book/security/configuring-content-security-policy/>`__ some features of an interactive log
+might not work properly, and its contents might be not shown or displayed incorrectly.
+
+A recommended way to allow Jenkins server to render interactive user content is to `configure Jenkins to publish
+artifacts on a different domain <https://www.jenkins.io/doc/book/security/user-content/#resource-root-url>`__
+by changing ``Resource Root URL`` in `Manage Jenkins » Configure System » Serve resource files from another domain`
+from something like ``my.jenkins.io`` to ``res.my.jenkins.io``.
+
+Note that Jenkins interaction with resource domain, resolved to the same host IP is not done via ``localhost``
```
i-keliukh, line 79: I think you can mention that jenkins requires some initial interaction with resource domain during changing of this setting.

doc/jenkins.rst
```diff
@@ -0,0 +1,136 @@
...
+------------------------
+
+As already mentioned above, due to `Jenkins Content-Security-Policy
+<https://www.jenkins.io/doc/book/security/configuring-content-security-policy/>`__ some features of an interactive log
+might not work properly, and its contents might be not shown or displayed incorrectly.
+
+A recommended way to allow Jenkins server to render interactive user content is to `configure Jenkins to publish
+artifacts on a different domain <https://www.jenkins.io/doc/book/security/user-content/#resource-root-url>`__
+by changing ``Resource Root URL`` in `Manage Jenkins » Configure System » Serve resource files from another domain`
+from something like ``my.jenkins.io`` to ``res.my.jenkins.io``.
```
i-keliukh, line 77: Can we use some placeholders for domain names instead of real ones?

doc/jenkins.rst
```diff
@@ -0,0 +1,138 @@
...
+Note that this step is only needed if `Nginx reverse proxy
+<https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/>`__ is used.
+
+To understand why these fixes are needed, let's pay more attention to the mechanism of 'another domain', used by
+Jenkins. When requesting an artifact, that is served from another domain, user first goes to main Jenkins web
+server, that returns a redirection link to acquire a said artifact.
+
+As `specified in docs <https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/#passing-request-headers>`__,
+without specification Nginx replaces ``Host`` headers with ``$proxy_host``. In this case it changes
+``res.my.jenkins.io`` to proxy IP and port specifications. The problem is, when received such specifications,
```
i-keliukh, line 122: 1. Please change res.my.jenkins.io -> <resource domain>
2. is port really passed as a part of 'Host' header?

doc/jenkins.rst
```diff
@@ -0,0 +1,138 @@
...
+Jenkins server no longer understands the request is sent to resource domain, and returns the ``404 NOT FOUND`` error.
```
i-keliukh, line 123: I propose to rephrase like this:
The problem is that without the ``Host`` header the Jenkins server is not able to understand that the request is sent to the resource domain, and returns the ``404 NOT FOUND`` error.

Optionally you can split this sentence into two.

doc/jenkins.rst
```diff
@@ -0,0 +1,138 @@
...
+
+Note that this step is only needed if `Nginx reverse proxy
+<https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/>`__ is used.
+
+To understand why these fixes are needed, let's pay more attention to the mechanism of 'another domain', used by
+Jenkins. When requesting an artifact, that is served from another domain, user first goes to main Jenkins web
+server, that returns a redirection link to acquire a said artifact.
+
+As `specified in docs <https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/#passing-request-headers>`__,
+without specification Nginx replaces ``Host`` headers with ``$proxy_host``. In this case it changes
```
i-keliukh, line 121: Please change "headers" to "header".

doc/jenkins.rst
```diff
@@ -0,0 +1,138 @@
...
+    location / {
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+    }
+
+so that real headers are passed to Jenkins to handle the resource domain magic.
+
+.. note::
+
+    Also you might need to set ``server_name`` to ``my.jenkins.io res.my.jenkins.io`` if they are located
```
i-keliukh, line 137: The difference between one or two server entries is is defined by the difference of their external IP addresses.
If the physical server is the same and the external IP address is the same, both can be described by a single entry within nginx config.

If there are two different IPs, two options are possible:
1. they are still hosted on the same machine -> a single nginx instance can server the (single config file), but two separate entries are required
2. they are hosted on two different machines -> two nginx instances are requied -> two separate config files

I propose to remove "same host" and related parts from this sentence.

doc/jenkins.rst
```diff
@@ -118,9 +118,10 @@ Jenkins. When requesting an artifact, that is served from another domain, user f
 server, that returns a redirection link to acquire a said artifact.
 
 As `specified in docs <https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/#passing-request-headers>`__,
-without specification Nginx replaces ``Host`` headers with ``$proxy_host``. In this case it changes
-``res.my.jenkins.io`` to proxy IP and port specifications. The problem is, when received such specifications,
-Jenkins server no longer understands the request is sent to resource domain, and returns the ``404 NOT FOUND`` error.
+without specification Nginx replaces ``Host`` header with ``$proxy_host``. In this case it changes
+``<resource domain>`` to proxy IP specifications.The problem is that without the Host header the Jenkins server
```
i-keliukh, line 122: Space after dot

doc/jenkins.rst
```diff
@@ -118,9 +118,10 @@ Jenkins. When requesting an artifact, that is served from another domain, user f
...
+is not able to understand that the request is sent to the resource domain (and not the main one). Therefore it returns
```
i-keliukh, line 123: I think "and not the main one" is incorrect assumption.
I believe the jenkins server interprets all requests without the 'Host' header as targeting the main domain.
However, in this context this detail is not important.

i-keliukh (approved):

PR #694, opened 2021-12-02
feat(output): wrap links in HTML log

miltolstoy:


universum/modules/output/html_output.py
```diff
@@ -128,6 +130,19 @@ def _build_html_head(self):
             head.append(f"<style>{css_file.read()}</style>")
         return "".join(head)
 
+    @staticmethod
+    def _wrap_links(line):
+        position_shift = 0
+        pattern = r"(http|https|ftp|file|mailto):\S+"
```
i-keliukh, line 136: I suggest to change just "\S+" to something like "(\S|\\\ )+" to match escaped spaces within url.

i-keliukh (approved):

k-dovgan:
Looks good to me.

PR #698, opened 2021-12-17
fix(mypy): if a variable can be set to None, the typing should be 'Optional'

k-dovgan:


miltolstoy (approved):

universum/modules/error_state.py
```diff
@@ -36,7 +36,7 @@ def check_required_option(self, setting_name: str, message: str) -> bool:
         return True
 
     def read_multiline_option(self, setting_name: str) -> str:
-        value: str = getattr(self.settings, setting_name, None)
+        value: str = getattr(self.settings, setting_name, "")
```
i-keliukh, line 39: The check in the next line is not needed anymore

i-keliukh (approved):

universum/modules/error_state.py
```diff
@@ -36,17 +36,17 @@ def check_required_option(self, setting_name: str, message: str) -> bool:
...
-                return ""
-        return value
+        if not value.startswith('@'):
+            return value
+        try:
+            with open(value.lstrip('@'), encoding="utf-8") as value_file:
+                return value_file.read()
+        except FileNotFoundError as e:
+            self.error(f"Error reading argument {setting_name} from file {e.filename}: no such file")
+            return ""
```
i-keliukh, line 49: I don't know why, but I prefer to have this return statement on the function level, not inside except block

i-keliukh (approved):

PR #705, opened 2022-02-08
Add Github Actions output support

dogbert911:
# Description

Add Github Actions out type support.

The main goal of these changes is to add support for grouping logs to the Github Action.
At the moment, github does not support nested logs, so the beginning of each new block of logs closes the previous one. 
If the block ends with an error, a error message is displayed.

Resolves #(issue) None

# How Has This Been Tested?

See result on Github Actions:
https://github.com/dogbert911/universum_test/runs/5246061477?check_suite_focus=true

# Checklist:

- [X] My code follows the [PEP 8](https://www.python.org/dev/peps/pep-0008/)
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] New and existing tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/universum/lib/utils.py
+++ b/universum/lib/utils.py
@@ -57,29 +57,35 @@ def detect_environment() -> str:
     """
     :return: "tc" if the script is launched on TeamCity agent,
              "jenkins" is launched on Jenkins agent,
+             "github" is launched on Github Actions agent,
              "terminal" otherwise
     """
     teamcity = "TEAMCITY_VERSION" in os.environ
     jenkins = "JENKINS_HOME" in os.environ
     pycharm = "PYCHARM_HOSTED" in os.environ
+    github = "GITHUB_WORKFLOW" in os.environ
     if pycharm:
         return "terminal"
     if teamcity and not jenkins:
         return "tc"
     if not teamcity and jenkins:
         return "jenkins"
+    if github:
+        return "github"
     return "terminal"
 
 
 LocalFactoryT = TypeVar('LocalFactoryT', bound=Module)
 TeamcityFactoryT = TypeVar('TeamcityFactoryT', bound=Module)
 JenkinsFactoryT = TypeVar('JenkinsFactoryT', bound=Module)
+GithubFactoryT = TypeVar('GithubFactoryT', bound=Module)
 
 
 def create_driver(local_factory: Callable[[], LocalFactoryT],
                   teamcity_factory: Callable[[], TeamcityFactoryT],
                   jenkins_factory: Callable[[], JenkinsFactoryT],
-                  env_type: str = "") -> Union[LocalFactoryT, TeamcityFactoryT, JenkinsFactoryT]:
+                  github_factory:  Callable[[], GithubFactoryT],
+                  env_type: str = "") -> Union[LocalFactoryT, TeamcityFactoryT, JenkinsFactoryT, GithubFactoryT]:
     if not env_type:
         env_type = detect_environment()
 
@@ -87,6 +93,8 @@ def create_driver(local_factory: Callable[[], LocalFactoryT],
         return teamcity_factory()
     if env_type == "jenkins":
         return jenkins_factory()
+    if env_type == "github":
+        return github_factory()
     return local_factory()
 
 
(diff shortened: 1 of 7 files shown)


dogbert911:
Upd: https://github.com/dogbert911/universum_test/runs/5152964959?check_suite_focus=true

universum/modules/output/github_output.py
```diff
@@ -0,0 +1,61 @@
...
+
+__all__ = [
+    "GithubOutput"
+]
+
+
+class GithubOutput(BaseOutput):
+    """
+    GitHub doesn't support nested grouping
+    See: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines
```
i-keliukh, line 11: I think the more correct link would be https://github.com/actions/runner/issues/802

dogbert911: done

universum/modules/output/output.py
```diff
@@ -21,14 +22,15 @@ class Output(Module):
...
     def define_arguments(argument_parser):
         parser = argument_parser.get_or_create_group("Output", "Log appearance parameters")
-        parser.add_argument("--out-type", "-ot", dest="type", choices=["tc", "term", "jenkins"],
-                            help="Type of output to produce (tc - TeamCity, jenkins - Jenkins, term - terminal). "
-                                 "TeamCity and Jenkins environments are detected automatically "
-                                 "when launched on build agent.")
+        parser.add_argument("--out-type", "-ot", dest="type", choices=["tc", "term", "jenkins", "github"],
+                            help="Type of output to produce (tc - TeamCity, jenkins - Jenkins, term - terminal, "
+                                 "github - Github Actions). TeamCity and Jenkins environments are detected "
+                                 "automatically when launched on build agent.")
```
k-dovgan, line 33: You need to add a comment whether GitHub is detected automatically, or should this output type be always triggered by conscious user action (and if latter, then what action exactly: like, setting up a special environment variable)

universum/lib/utils.py
```diff
@@ -57,36 +57,44 @@ def detect_environment() -> str:
     """
     :return: "tc" if the script is launched on TeamCity agent,
              "jenkins" is launched on Jenkins agent,
+             "github" is launched on Github Actions agent,
```
k-dovgan, line 60: I'm not sure "GitHub Actions agent" is a correct statement. Do they use "agents" as well?

universum/modules/output/github_output.py
```diff
@@ -0,0 +1,47 @@
...
+
+__all__ = [
+    "GithubOutput"
+]
+
+
+class GithubOutput(TerminalBasedOutput):
+    """
+    GitHub doesn't support nested grouping
+    See: https://github.com/actions/runner/issues/802
```
k-dovgan, line 11: I think, if you did use some official user guide for GitHub Actions, it would be nice to add the link here too, not only that one to a very specific issue.

universum/modules/output/github_output.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._block_opened = False
+
+    # pylint: disable = arguments-differ
```
k-dovgan, line 18: Why do we silence this issue instead of fixing it?

universum/modules/output/github_output.py
```diff
@@ -0,0 +1,47 @@
...
+        self.print_lines(f"::group::{num_str} {name}")
+        self._block_opened = True
+
+    def close_block(self, num_str, name, status):
+        if self._block_opened:
+            self._block_opened = False
+            self.print_lines("::endgroup::")
+
+        if status == "Failed":
+            self.print_lines(f'::error::{num_str} {name} - Failed')
```
k-dovgan, line 38: Not a serious problem, but you use double quotes everywhere except for this line (so it looks kinda strange).

universum/modules/output/terminal_based_output.py
```diff
@@ -4,6 +4,7 @@
 from .base_output import BaseOutput
 
 __all__ = [
+    "stdout",
```
k-dovgan, line 7: How about moving this function to 'BaseOutput', for example? It is used by two different but architecturally equal classes.

k-dovgan (approved):

PR #706, opened 2022-02-15
fix: throw an error if project config is empty

miltolstoy:
**Before**: empty config produces no error, but Universum did nothing.
```
3. Executing build steps
 └ [Success]
```

**Now**: throwing an exception if project configs are empty.
```
2. Processing project configs
 |   ==> Adding file /home/alelyak/github/Universum/tests/artifacts/CONFIGS_DUMP.txt to artifacts...
 └ [Success]

==> Project configs are empty, build skipped
 
3. Finalizing
 |   3.1. Cleaning copied sources
 |      └ [Success]
 |   
 └ [Success]
```

### DIFF
--- a/tests/test_run_steps_filter.py
+++ b/tests/test_run_steps_filter.py
@@ -1,6 +1,8 @@
 import pytest
 
+from universum import __main__
 from .deployment_utils import UniversumRunner
+
 
 config = """
 from universum.configuration_support import Configuration
@@ -10,6 +12,11 @@ def step(name, cmd=False):
 
 configs = step('parent 1 ') * (step('step 1', True) + step('step 2', True))
 configs += step('parent 2 ') * (step('step 1', True) + step('step 2', True))
+"""
+
+empty_config = """
+from universum.configuration_support import Configuration
+configs = Configuration()
 """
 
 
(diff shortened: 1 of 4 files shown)


k-dovgan:
This is a test comment.

universum/main.py
```diff
@@ -72,6 +72,10 @@ def execute(self) -> None:
 
         self.vcs.prepare_repository()
         project_configs: Configuration = self.launcher.process_project_configs()
+        if not project_configs:
```
i-keliukh, line 75: The issues with this approach are the following:
1. Nonci mode is not considered. It could be intentional or not, but the intention is not expressed in the source code.
2. Logically configs should be handled by launcher, not main
3. Because of item 2, the SilentAbortException is used instead of CriticalCiException. The latter is better choice in general case.

tests/test_run_steps_filter.py
```diff
@@ -1,7 +1,9 @@
 import pytest
 
+from universum import __main__
```
k-dovgan, line 3: Should we change the other test in this suite to running '__main__' instead of docker too? Maybe later in a separate PR?

k-dovgan (approved):

i-keliukh (approved):

PR #708, opened 2022-03-08
test: remove docker usage from filter tests

miltolstoy:


tests/test_run_steps_filter.py
```diff
@@ -19,75 +19,66 @@ def step(name, cmd=False):
...
+    captured = capsys.readouterr()
 
-def test_config_empty_nonci(tmpdir, capsys):
-    check_empty_config_error(tmpdir, capsys, nonci_cli_params)
+    assert return_code == 0
+    for log_str in expected_logs:
+        assert log_str in captured.out
+    for log_str in unexpected_logs:
+        assert log_str not in captured.out
+    assert not captured.err
```
k-dovgan, line 59: I can't find a place where we expect stderr and check its contents. Maybe, using FuzzyCallChecker would suite these tests better?

k-dovgan (approved):

PR #710, opened 2022-03-17
feat(structure_handler): initialize conditional steps logic

miltolstoy:
Initial change for the conditional steps logic, the only basic scenario is implemented and tested. See a newly added test file for a usage example.

Additional logic to handle in the following PRs (which I'm aware of):
- steps addition and multiplying
- steps filtering
- the "critical" flag
- the "if_env_set" flag
- background execution

Please update this list with any ideas to check/implement.

### DIFF
--- a/tests/test_conditional_steps.py
+++ b/tests/test_conditional_steps.py
@@ -1,11 +1,12 @@
 import re
 import pytest
+import inspect
 
 from universum import __main__
 
 
-true_branch_step_name = "true branch"
-false_branch_step_name = "false branch"
+true_branch_step_name = "true_branch"
+false_branch_step_name = "false_branch"
 
 
 def test_conditional_true_branch(tmpdir, capsys):
@@ -23,13 +24,19 @@ def check_conditional_step_success(tmpdir, capsys, conditional_step_passed):
 
 def build_config_file(tmpdir, conditional_step_passed):
     conditional_step_exit_code = 0 if conditional_step_passed else 1
-    config = "from universum.configuration_support import Configuration, Step\n"
-    config += f"true_branch_step = Step(name='{true_branch_step_name}', command=['pwd'])\n"
-    config += f"false_branch_step = Step(name='{false_branch_step_name}', command=['ls'])\n"
-    config += "conditional_step = Configuration([{'name': 'conditional',\n"
-    config += f"   'command': ['bash', '-c', 'exit {conditional_step_exit_code}'],\n"
-    config += "    'if_succeeded': true_branch_step, 'if_failed': false_branch_step}])\n"
-    config += "configs = conditional_step"
+
+    config = inspect.cleandoc(f'''
+        from universum.configuration_support import Configuration, Step
+
+        true_branch_step = Step(name='{true_branch_step_name}', command=['touch', '{true_branch_step_name}'],
+                                artifacts='{true_branch_step_name}')
+        false_branch_step = Step(name='{false_branch_step_name}', command=['touch', '{false_branch_step_name}'])
+        conditional_step = Configuration([dict(name='conditional',
+            command=['bash', '-c', 'exit {conditional_step_exit_code}'],
+            if_succeeded=true_branch_step, if_failed=false_branch_step)])
+
+        configs = conditional_step
+    ''')
 
     config_file = tmpdir.join("configs.py")
     config_file.write_text(config, "utf-8")
@@ -48,6 +55,7 @@ def check_conditional_step(tmpdir, capsys, config_file, conditional_step_passed)
     assert return_code == 0
 
     captured = capsys.readouterr()
+    print(captured.out)
     conditional_succeeded_regexp = r"conditional.*Success.*\|   5\.2"
     assert re.search(conditional_succeeded_regexp, captured.out, re.DOTALL)
 
--- a/tests/test_conditional_steps.py
+++ b/tests/test_conditional_steps.py
@@ -28,8 +28,7 @@ def build_config_file(tmpdir, conditional_step_passed):
     config = inspect.cleandoc(f'''
         from universum.configuration_support import Configuration, Step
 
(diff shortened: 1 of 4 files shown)


tests/test_conditional_steps.py
```diff
@@ -0,0 +1,57 @@
...
+
+def check_conditional_step_success(tmpdir, capsys, conditional_step_passed):
+    config_file = build_config_file(tmpdir, conditional_step_passed)
+    check_conditional_step(tmpdir, capsys, config_file, conditional_step_passed)
+
+
+def build_config_file(tmpdir, conditional_step_passed):
+    conditional_step_exit_code = 0 if conditional_step_passed else 1
+    config = "from universum.configuration_support import Configuration, Step\n"
+    config += f"true_branch_step = Step(name='{true_branch_step_name}', command=['pwd'])\n"
```
k-dovgan, line 27: Please use triple quotes instead of line adding for better readability

i-keliukh: For even better readability I suggest to use `inspect.cleandoc`

universum/modules/structure_handler.py
```diff
@@ -230,6 +252,12 @@ def execute_step_structure(self, configs: Configuration, step_executor) -> None:
         if self.active_background_steps:
             self.run_in_block(self.report_background_steps, "Reporting background steps", False)
 
+    def _build_next_step_name(self, name):
```
i-keliukh, line 255: The name of the function suggests that it doesn't have side effects.
Either the name or the logic has to be fixed.

universum/modules/structure_handler.py
```diff
@@ -221,6 +240,9 @@ def report_background_steps(self) -> bool:
 
     def execute_step_structure(self, configs: Configuration, step_executor) -> None:
         self.configs_total_count = sum(1 for _ in configs.all())
+        for config in configs.all():
+            if config.if_succeeded or config.if_failed:
+                self.configs_total_count += 1
```
i-keliukh, line 245: I think you can just merge this calculation to with the comprehension above.
Something like:
```python
sum(2 if x.if_succeeded or x.if_failed else 1 for x in configs.all())
```

Not sure this would be better, but I prefer the calculation to be uniform.
You can merge the above calculation with your for loop instead.

universum/modules/structure_handler.py
```diff
@@ -181,10 +187,23 @@ def execute_steps_recursively(self, parent: Optional[Step], cfg: Configuration,
...
+                elif is_conditional:
+                    try:
+                        step_name = self._build_next_step_name(item.name)
+                        self.run_in_block(self.execute_one_step, step_name, False,
+                                          item, step_executor, obj_a.critical)
+                        if obj_a.if_succeeded:
+                            step_name = self._build_next_step_name(obj_a.if_succeeded.name)
+                            self.run_in_block(self.execute_one_step, step_name, False,
+                                              obj_a.if_succeeded, step_executor, obj_a.critical)
+                    except (SilentAbortException, StepException):  # user-defined step failed
```
i-keliukh, line 199: I am not sure you need to catch SilentAbortException here.

tests/test_conditional_steps.py
```diff
@@ -0,0 +1,57 @@
...
+              "--clean-build",
+              "-o", "console"]
+    params.extend(["-cfg", str(config_file)])
+
+    return_code = __main__.main(params)
+    assert return_code == 0
+
+    captured = capsys.readouterr()
+    conditional_succeeded_regexp = r"conditional.*Success.*\|   5\.2"
+    assert re.search(conditional_succeeded_regexp, captured.out, re.DOTALL)
```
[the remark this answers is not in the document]
miltolstoy: For now, artifacts collection doesn't work in the "branch" steps. I'll fix it in one of the next PRs and extend these checks then.

tests/test_conditional_steps.py
```diff
@@ -0,0 +1,57 @@
...
+
+    expected_log = true_branch_step_name if conditional_step_passed else false_branch_step_name
+    unexpected_log = false_branch_step_name if conditional_step_passed else true_branch_step_name
+    assert expected_log in captured.out
+    assert not unexpected_log in captured
```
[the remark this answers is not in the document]
miltolstoy: No, we don't expect a strderr. But I need a whole log to validate step status, and FuzzyCallChecker can check it line-by-line.
Maybe later I will update the FuzzyCallChecker and this test will use it.

k-dovgan (dismissed):
I like this implementation a lot more than previous one.

universum/modules/structure_handler.py
```diff
@@ -140,7 +140,12 @@ def execute_one_step(self, configuration: Step,
 
         process.start()
         if not configuration.background:
-            process.finalize()
+            try:
+                process.finalize()
+            except StepException:
+                if configuration.if_failed:  # conditional step with next step configured at `if_failed` event
```
i-keliukh, line 146: This is a generic function for executing single steps.
Could you please consider moving this conditional-steps-specific functionality to the respective functions?

universum/modules/structure_handler.py
```diff
@@ -208,6 +212,26 @@ def execute_steps_recursively(self, parent: Optional[Step], cfg: Configuration,
...
+
+    def execute_conditional_step(self, step, step_executor):
+        try:
+            self.execute_one_step(step, step_executor, step.critical)
+            self.close_block()
+            if step.if_succeeded:
+                self.execute_conditional_branch_step(step.if_succeeded, step_executor)
+        except StepException:
+            if step.if_failed:
+                self.close_block()
```
i-keliukh, line 224: Is there alternative to manual opening and closing blocks?
In my opinion it is better to have the same structure of blocks in code and in structure. This is what `run_in_block` function for.

i-keliukh: If you really want to commit this code, please go ahead. But please also consider immediately refactoring it.

i-keliukh (approved):

k-dovgan (approved):

Issue #703, opened 2022-02-02
Command line options '--p4-mappings' work weird


dogbert911:
## Description

When using the options `--p4-mappings '//my/mapping/... /mapping/...'`  universum return next error:
```
 |   2.2. Creating workspace
 |      |   2.2.1. Cleaning workspace
 |      |        |   ==> No client to delete
 |      |        └ [Success]
 |      |   
 |      |   Error: [P4#run] Errors during command execution( "p4 client -i" )
 |      |   
 |      |   	[Error]: "Error in client specification.\nMapping 'default' is not under '//WATCH_TASK/...'."
 |      |   
 |      └ [Failed]

```
But if this option is used as  `--p4-mappings='//my/mapping/... /mapping/...'` all right.

## Steps to reproduce

Run  in terminal:
```
python -m universum --vcs-type p4 --p4-client MY_CLIENT --p4-port 111.111.111.111:111 --p4-user USER --p4-password PASS --p4-mappings '//my/mapping/... /mapping/...'
```
## Expected result

```
 |   2.2. Creating workspace
 |      └ [Success]

```

## Actual result
```
 |   2.2. Creating workspace
 |      |   2.2.1. Cleaning workspace
 |      |        |   ==> No client to delete
 |      |        └ [Success]
 |      |   
 |      |   Error: [P4#run] Errors during command execution( "p4 client -i" )
 |      |   
 |      |   	[Error]: "Error in client specification.\nMapping 'default' is not under '//WATCH_TASK/...'."
 |      |   
 |      └ [Failed]

```

## Environment

 - Python version: 3.8.10
 - Universum version: 0.19.12

## Additional context

None

PR #713, opened 2022-04-02
fix(Output): use default HTML log name if '-hl' CLI parameter was last

### LINKED ISSUE #711: Wrong default name of html log
Currently when using the `-hl` option without any further specification the name of a created file is `default` (instead of expected `universum_log.html`).
Need to be expected, fixed and, maybe, covered by regression test

miltolstoy:
The argument parser works differently depending on a CLI parameter position.
`universum -hl ${other_params}` -> `settings.html_log == "universum_log.html"`
`universum ${other_params} -hl` -> `settings.html_log == "default"`

Fix this to have a proper HTML log name regardless of CLI parameter position.
Resolves #711

### DIFF
--- a/tests/test_html_output.py
+++ b/tests/test_html_output.py
@@ -74,6 +74,14 @@ def test_cli_log_custom_name(tmpdir):
 
 def test_cli_log_default_name(tmpdir):
     artifact_dir = check_cli(tmpdir, ["-hl"])
+    assert not os.path.exists(os.path.join(artifact_dir, "default"))
+    assert os.path.exists(os.path.join(artifact_dir, "universum_log.html"))
+
+
+# https://github.com/Samsung/Universum/issues/711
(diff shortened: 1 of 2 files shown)


i-keliukh:
In the description the "Output" should be lowercase.
However, this doesn't really matter because the fix should be generic to fix both issues #703 and #711.

i-keliukh:
The real source of these bugs is located in `_add_default_parser` function in https://github.com/Samsung/Universum/blob/96cb237b1e11cf835d86779e4f960b284ac3f720/universum/lib/module_arguments.py#L136-L147

miltolstoy:
Dropping this review because a more general fix is needed.
The same issue exists for other CLI arguments as well - https://github.com/Samsung/Universum/issues/703. A more high-level CLI parser fix is required.

Issue #709, opened 2022-03-14
Implement conditional steps execution


miltolstoy:
Desired scenario:
0. Three steps exist - "conditional" step, "true branch" step, "false branch" step.
1. Execute the "conditional" step.
2.1. If the "conditional" step succeeded, execute the "true branch" step
2.2. If the "conditional" step failed, mark it as succeeded and execute the "false branch" step

See discussions and proposals under https://github.com/Samsung/Universum/issues/636

miltolstoy:
Preliminary development plan was created during the task discussion:
1. Step(..., if_succeeded=..., if_failed=...)
2. ? Step -> If / IfStep / ConditionalStep
3. ? ConditionalStep(...).then(...).else(...)
4. ? conditional_step >> then | else

"?" steps are subject to discussion.

i-keliukh:
1. Children
> handle config where conditional step has children
Note that there are several separate design decisions:
- how to deal with children in conditional step itself
- what to do with children in "true branch" and "false branch"
- do we have to put "true branch" and "false branch" into the same structure as conditional step itself?

2. TeamCity tags requirements are missing

PR #716, opened 2022-04-27
feat(artifact_collector): artifacts collection for conditional steps

miltolstoy:
This PR contains:
1. Handling conditional step branches artifacts
2. Conditional steps tests update to check files created instead of the log line

Similar changes will be applied to the `configuration.report_artifacts` in the next PR after this change will be reviewed and updated accordingly.

Relates #709

### DIFF
--- a/tests/test_conditional_steps.py
+++ b/tests/test_conditional_steps.py
@@ -3,51 +3,59 @@ import os
 import inspect
 
 from universum import __main__
-from universum.configuration_support import Step, Configuration
+from universum.configuration_support import Step
 
 
-conditional_step_name = "conditional"
-true_branch_step_name = "true_branch"
-false_branch_step_name = "false_branch"
+class StepsInfo:
+    conditional_step = None
+    true_branch_step = None
+    false_branch_step = None
+    is_conditional_step_passed = False
 
 
 def test_conditional_true_branch(tmpdir, capsys):
-    check_conditional_step_success(tmpdir, capsys, conditional_step_passed=True)
+    steps_info = get_conditional_steps_info(is_conditional_step_passed=True)
+    check_conditional_step(tmpdir, capsys, steps_info)
 
 
 def test_conditional_false_branch(tmpdir, capsys):
-    check_conditional_step_success(tmpdir, capsys, conditional_step_passed=False)
+    steps_info = get_conditional_steps_info(is_conditional_step_passed=False)
+    check_conditional_step(tmpdir, capsys, steps_info)
 
 
-def check_conditional_step_success(tmpdir, capsys, conditional_step_passed):
-    config_file = build_config_file(tmpdir, conditional_step_passed)
-    check_conditional_step(tmpdir, capsys, config_file, conditional_step_passed)
+def get_conditional_steps_info(is_conditional_step_passed):
+    steps_info = StepsInfo()
 
+    conditional_step_name = "conditional"
+    conditional_step_exit_code = 0 if is_conditional_step_passed else 1
+    steps_info.conditional_step = Step(
+        name=conditional_step_name,
+        command=["bash", "-c", f"touch {conditional_step_name}; exit {conditional_step_exit_code}"],
+        artifacts=conditional_step_name)
+    steps_info.is_conditional_step_passed = is_conditional_step_passed
 
-def build_config_file(tmpdir, conditional_step_passed):
-    conditional_step_exit_code = 0 if conditional_step_passed else 1
-
-    true_branch_step = Step(
+    true_branch_step_name = "true_branch"
+    steps_info.true_branch_step = Step(
         name=true_branch_step_name,
-        command=['touch', true_branch_step_name],
+        command=["touch", true_branch_step_name],
         artifacts=true_branch_step_name)
 
-    false_branch_step = Step(
+    false_branch_step_name = "false_branch"
+    steps_info.false_branch_step = Step(
         name=false_branch_step_name,
-        command=['touch', false_branch_step_name],
+        command=["touch", false_branch_step_name],
         artifacts=false_branch_step_name)
 
-    conditional_step = Step(
-        name=conditional_step_name,
-        command=['bash', '-c', f'touch {conditional_step_name}; exit {conditional_step_exit_code}'],
-        artifacts=conditional_step_name)
+    return steps_info
 
+
+def write_config_file(tmpdir, conditional_steps_info):
     config = inspect.cleandoc(f'''
         from universum.configuration_support import Configuration, Step
 
-        true_branch_step = Step(**{str(true_branch_step)})
-        false_branch_step = Step(**{str(false_branch_step)})
-        conditional_step = Step(**{str(conditional_step)})
+        true_branch_step = Step(**{str(conditional_steps_info.true_branch_step)})
+        false_branch_step = Step(**{str(conditional_steps_info.false_branch_step)})
+        conditional_step = Step(**{str(conditional_steps_info.conditional_step)})
         
         # `true/false_branch_steps` should be Python objects from this script
         conditional_step.is_conditional = True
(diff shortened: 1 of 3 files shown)


universum/modules/artifact_collector.py
```diff
@@ -128,48 +139,61 @@ def preprocess_artifact_list(self, artifact_list, ignore_already_existing=False)
...
         self.html_output.artifact_dir_ready = True
         artifact_list = []
         report_artifact_list = []
         for configuration in project_configs.all():
             if configuration.artifacts:
-                path = utils.parse_path(configuration.artifacts, self.settings.project_root)
-                artifact_list.append(dict(path=path, clean=configuration.artifact_prebuild_clean))
+                artifact_list.append(self.get_config_artifact(configuration))
+            if configuration.is_conditional:
+                artifacts_info = self.get_conditional_step_branches_artifacts(configuration)
```
i-keliukh, line 194: I think you could implement the artifact support by adding artifacts of both branches to the list. All other code should not depend on the steps structure and execution status when dealing with artifacts.
Fortunately or unfortunately this is how artifacts are currently processed (if I am not mistaken).

universum/modules/artifact_collector.py
```diff
@@ -184,6 +208,32 @@ def set_and_clean_artifacts(self, project_configs: Configuration, ignore_existin
                                                                     name, True, report_artifact_list,
                                                                     ignore_existing_artifacts)
 
+    def get_conditional_step_branches_artifacts(self, configuration):
+        """
+        Both branches have artifacts -> return ConditionalStepArtifacts(dict, dict)
+        One branch have artifact -> return dict
```
i-keliukh, line 214: have -> has

universum/modules/artifact_collector.py
```diff
@@ -228,8 +278,30 @@ def collect_artifacts(self):
...
+    def collect_conditional_step_artifacts(self, artifacts_info):
+        succeeded_artifact_path = artifacts_info.succeeded_config_artifact["path"]
+        succeeded_artifact_exists = os.path.exists(succeeded_artifact_path)
+        failed_artifact_path = artifacts_info.failed_config_artifact["path"]
+        failed_artifact_exists = os.path.exists(failed_artifact_path)
+
+        # At this point we don't know which branch step was executed,
+        # so assuming that any single branch artifact presence is good
+        if succeeded_artifact_exists and failed_artifact_exists:
+            raise CriticalCiException("Both conditional step branches artifacts exist")
```
i-keliukh, line 296: I think this logic is incorrect.
Generally we allow the same artifact paths to be specified multiple times in different steps. 
If some regular step has artifact path "a" (and creates corresponding artifact) and skipped conditional step has the same artifact path "a", this line would trigger the exception, which is incorrect.

i-keliukh:
The requirements for the artifact processing for conditional steps are missing in this PR.
I think it would be good to have them listed in the form of product documentation included in this PR.

k-dovgan:
Please ignore this comment.

universum/modules/artifact_collector.py
miltolstoy: This code became redundant after the artifacts collection moment was changed from the end of the build to the end of each step.
Removed.

tests/test_conditional_steps.py
```diff
@@ -25,16 +27,34 @@ def check_conditional_step_success(tmpdir, capsys, conditional_step_passed):
...
+
     config = inspect.cleandoc(f'''
         from universum.configuration_support import Configuration, Step
 
-        true_branch_step = Step(name='{true_branch_step_name}', command=['touch', '{true_branch_step_name}'])
-        false_branch_step = Step(name='{false_branch_step_name}', command=['touch', '{false_branch_step_name}'])
-        conditional_step = Configuration([dict(name='conditional',
-            command=['bash', '-c', 'exit {conditional_step_exit_code}'],
-            if_succeeded=true_branch_step, if_failed=false_branch_step)])
+        true_branch_step = Step(**{str(true_branch_step)})
```
i-keliukh, line 48: It looks like this variable is not needed. But I think you can keep it for readability.

tests/test_conditional_steps.py
```diff
@@ -56,11 +107,20 @@ def check_conditional_step(tmpdir, capsys, config_file, conditional_step_passed)
...
+
+    assert os.path.exists(os.path.join(artifacts_dir, conditional_step_artifact))
+
+    expected_artifact = true_branch_step_artifact if is_conditional_step_passed else false_branch_step_artifact
+    if expected_artifact:
+        assert os.path.exists(os.path.join(artifacts_dir, expected_artifact))
+
+    unexpected_artifact = false_branch_step_artifact if is_conditional_step_passed else true_branch_step_artifact
+    if unexpected_artifact:
+        assert not os.path.exists(os.path.join(artifacts_dir, unexpected_artifact))
```
i-keliukh, line 126: No empty line at the end of file

tests/test_conditional_steps.py
```diff
@@ -1,40 +1,89 @@
...
+    conditional_step_artifact = steps_info.conditional_step["artifacts"]
+    steps_info.true_branch_step.command = ["touch", conditional_step_artifact]
+    steps_info.true_branch_step.artifacts = conditional_step_artifact
 
-def build_config_file(tmpdir, conditional_step_passed):
-    conditional_step_exit_code = 0 if conditional_step_passed else 1
+    check_conditional_step(tmpdir, capsys, steps_info)
 
+
+def test_no_artifact_in_executed_branch_step(tmpdir, capsys):
```
i-keliukh, line 39: Is the logic of this test to check that artifacts are not collected if they are not specified for step?
How does it relate to conditional steps? 
If it doesn't, I think this is a test for basic functionality of artifacts collection and should be a part of its tests.
If the idea is to test that artifacts are not collected for the opposing branch (fail for successful case and success for failure case), then the test is incorrect.

universum/modules/artifact_collector.py
```diff
@@ -175,6 +182,28 @@ def set_and_clean_artifacts(self, project_configs: Configuration, ignore_existin
             with self.structure.block(block_name=name, pass_errors=True):
                 self.preprocess_artifact_list(report_artifact_list, ignore_existing_artifacts)
 
+    def get_conditional_step_branches_artifacts(self, configuration: Step) -> List[ArtifactInfo]:
```
i-keliukh, line 185: Misleading parameter name: configuration for type Step.

universum/modules/artifact_collector.py
```diff
@@ -175,6 +182,28 @@ def set_and_clean_artifacts(self, project_configs: Configuration, ignore_existin
...
+        succeeded_config_artifact: Optional[ArtifactInfo] = self.get_config_artifact_if_exists(configuration.if_succeeded)
+        failed_config_artifact: Optional[ArtifactInfo] = self.get_config_artifact_if_exists(configuration.if_failed)
+
+        if succeeded_config_artifact and failed_config_artifact:
```
i-keliukh, line 189: Please take a look a this code:
```python
result = []
if succeeded_config_artifact: result.append(succeeded_config_artifact)
if failed_config_artifact: result.append(failed_config_artifact)
return result
```

What do you think about its readability?

i-keliukh: Alternatively, the entire function can be merged into `preprocess_artifact_list` by using similar logic

miltolstoy:
> The requirements for the artifact processing for conditional steps are missing in this PR. I think it would be good to have them listed in the form of product documentation included in this PR.

It appears that artifact processing for conditional steps is similar to regular steps. It makes more sense to describe conditional steps logic in general and make small notes about artifacts processing in several places.
Agreed to be done in a separate PR because it's not related to the artifacts processing only.

universum/modules/artifact_collector.py
miltolstoy: > What do you think about its readability?

Much better, thanks.)

i-keliukh (approved):

PR #720, opened 2022-07-04
feat(config): add path to configuration file to configuration_support

### LINKED ISSUE #719: Add project config path to configuration_support
Currently only 'get_project_root' function is available. This function is not very helpful when using Universum in nonci mode, as it can be launched from any directory, and project root will be current directory.
Using absolute path to config file might help dealing with such situations.

k-dovgan:
Resolves #719

### DIFF
--- a/universum/configuration_support.py
+++ b/universum/configuration_support.py
@@ -704,6 +704,7 @@ class Configuration:
 
 
 global_project_root = os.getcwd()
+global_config_path = None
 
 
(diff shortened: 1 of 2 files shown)


i-keliukh (approved):

a-osipov-toxa:
Looks good, thank you!

i-keliukh (approved):

PR #722, opened 2022-07-06
fix(test): skip long tests when testing manually

i-keliukh:
Introduced environment variable for distinguishing test runs launched from
nox. Added pytest marker for skipping certain tests if the current session
is not launched by nox (if the environment variable is not set).


miltolstoy (commented):
LGTM

k-dovgan (approved):

PR #725, opened 2022-07-18
fix(docs): add reference to filter usage in non-CI

### LINKED ISSUE #721: Update step_filter documentation
Here the description of $STEP_FILTER is present:
https://universum.readthedocs.io/en/latest/additional_commands.html?highlight=filter#run-universum-locally-non-ci-mode

And here is not:
https://universum.readthedocs.io/en/latest/args.html?highlight=filter#command-line-and-parameters

Investigate if documentation update is needed.

k-dovgan:
Resolves #721

### DIFF
--- a/doc/additional_commands.rst
+++ b/doc/additional_commands.rst
@@ -45,6 +45,9 @@ Universum in non-CI mode has the following differences from default mode:
     :prog: {python} -m universum
     :path: run
 
+    --filter -f : @after
+        Read more about using filters `in the main article about arguments. <args.html#Configuration\ execution>`__
+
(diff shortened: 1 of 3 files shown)


doc/additional_commands.rst
```diff
@@ -45,6 +45,9 @@ Universum in non-CI mode has the following differences from default mode:
     :prog: {python} -m universum
     :path: run
 
+    --filter -f : @after
```
i-keliukh, line 48: Can we place the text regarding filters into separate file and include it here and in the main documentation?

PR #726, opened 2022-07-18
refactor(structure): remove StepException

i-keliukh:
Main reason for the removal of StepException is that code flow logic should
not be based on exceptions. With this change, the structure handler keeps
step status explicitly in local variables.

Related changes:

1. Added context manager for opening and closing blocks. This improves
readability by splitting parameters to block opening function and the
called code. Eventually all calls to `run_in_block` should be replaced by
the code that uses context manager in the 'with' block.

2. Simplified `execute_steps_recursively` function.

3. Added failing of group (parent) blocks if some child step failed.
Previously in such case the group was not marked as failed in logs. At the
same time, if the group was critical, subsequent steps in its parent group
were all skipped. This change makes logging consistent with behavior.

4. Added failing of "waiting" block in case of the failure of the
background block it is waiting for. This is mostly cosmetic change, because
such "waiting" block are created dynamically and they do not directly
affect the status of the group they belong to.

5. Polished the output of skipped blocks by wrapping block names in single
quotes.

### DIFF
--- a/universum/configuration_support.py
+++ b/universum/configuration_support.py
@@ -358,7 +358,8 @@ class Step:
     def __add__(self, other: 'Step') -> 'Step':
         """
         This functions defines operator ``+`` for :class:`Step` class objects by
-        concatenating strings and contents of dictionaries. Note that `critical` attribute is not merged.
+        concatenating strings and contents of dictionaries.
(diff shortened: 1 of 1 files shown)


k-dovgan (approved):

o-andrieiev (commented):
LGTM

PR #718, opened 2022-05-09
refactor: collect step artifacts immediately after step finish

miltolstoy:
This change was requested for conditional steps artifacts collections easier development.
Relates #717

Logic before:
  1. Whole artifacts list preprocessing
  2. Steps execution
  3. Whole artifacts list collection

Logic now:
  1. Whole artifacts list preprocessing (same)
  2. For each step:
    2.1. Step execution
    2.2. Step's artifacts collection

Log before:
```
5. Executing build steps
 |   5.1.  [ 1/4 ] Step with artifacts
 |      |   ==> Adding file /home/alelyak/Desktop/universum_test/artifacts/Step_with_artifacts_log.txt to artifacts...
 |      |   ==> Execution log is redirected to file
 |      |   $ /usr/bin/touch artifact1
 |      └ [Success]
 |
 |   5.2.  [ 2/4 ] Step with no artifacts
 |      |   ==> Adding file /home/alelyak/Desktop/universum_test/artifacts/Step_with_no_artifacts_log.txt to artifacts...
 |      |   ==> Execution log is redirected to file
 |      |   ==> No 'command' found. Nothing to execute
 |      └ [Success]
 |
 |   5.3.  [ 3/4 ] Failed step with artifacts
 |      |   ==> Adding file /home/alelyak/Desktop/universum_test/artifacts/Failed_step_with_artifacts_log.txt to artifacts...
 |      |   ==> Execution log is redirected to file
 |      |   $ /usr/bin/bash -c 'touch artifact4; exit 1'
 |      |   Error: Module sh got exit code 1
 |      └ [Failed]
 |
 |   Critical step failed. All further configurations will be skipped
 |   5.4. [ 4/4 ] 'Skipped step' skipped because of critical step failure
 └ [Failed]

6. Collecting artifacts
 |   6.1. Collecting 'artifact4'
 |      |   ==> Processing '/home/alelyak/Desktop/universum_test/temp/artifact4'
 |      └ [Success]
 |
 |   6.2. Collecting 'artifact1'
 |      |   ==> Processing '/home/alelyak/Desktop/universum_test/temp/artifact1'
 |      └ [Success]
 |
 └ [Success]

7. Reporting build result
 |   ==> Here is the summarized build result:
 |   ==> 5. Executing build steps
 |   ==>   5.1.  [ 1/4 ] Step with artifacts - Success
 |   ==>   5.2.  [ 2/4 ] Step with no artifacts - Success
[...]

PR #728, opened 2022-08-01
refactor(launcher): add RunningStep interface

miltolstoy:
Refactoring according to [this](https://github.com/Samsung/Universum/pull/718#discussion_r929894902) comment:

> Instead of returning tuple, I suggest to perform some refactoring:
> 
>     1. Declare interface for running step explicitly, as an abstract base class
> 
>     2. Use type declaration of this interface for type checking with mypy
> 
>     3. Include error status into this interface instead of returning it from the "start" and "finalize" functions separately
> 
>     4. Pass the object into active_background_steps instead of passing separate methods

### DIFF
--- a/universum/modules/launcher.py
+++ b/universum/modules/launcher.py
@@ -16,7 +16,7 @@ from ..lib.utils import make_block
 from . import automation_server, api_support, artifact_collector, reporter, code_report_collector
 from .output import HasOutput, Output
 from .project_directory import ProjectDirectory
-from .structure_handler import HasStructure
+from .structure_handler import HasStructure, RunningStepBase
 
 __all__ = [
     "Launcher",
@@ -158,19 +158,6 @@ def get_match_patterns(filters: Union[str, List[str]]) -> Tuple[List[str], List[
         elif f:
             include.append(f)
     return include, exclude
-
-
-class RunningStepBase(ABC):
-
-    error: Optional[str] = None
-
-    @abstractmethod
-    def start(self) -> None:
-        pass
-
-    @abstractmethod
-    def finalize(self):
-        pass
 
 
 class RunningStep(RunningStepBase):
--- a/universum/modules/launcher.py
+++ b/universum/modules/launcher.py
@@ -5,6 +5,7 @@ from inspect import cleandoc
 from typing import Callable, Dict, List, Optional, TextIO, Tuple, Union
 from requests import Response
 import sh
+from abc import ABC, abstractmethod
 
 from .error_state import HasErrorState
 from .. import configuration_support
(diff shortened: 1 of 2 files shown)


universum/modules/launcher.py
```diff
@@ -159,7 +160,20 @@ def get_match_patterns(filters: Union[str, List[str]]) -> Tuple[List[str], List[
     return include, exclude
 
 
-class RunningStep:
+class RunningStepBase(ABC):
```
axxie, line 163: Declare RunningStepBase in structure_handler.py and import it here

universum/modules/structure_handler.py
```diff
@@ -66,10 +67,23 @@ def is_successful(self) -> bool:
...
     name: str
     block: Block
-    finalizer: Callable[[], None]
+    process: RunningStepBase
     is_critical: bool
 
 
+class RunningStepBase(ABC):
+
+    error: Optional[str] = None
```
i-keliukh, line 76: The class and instance variables of abcs are not checked by mypy.
So it is better to use some explicit method, like 'get_error', for returning errors.

universum/modules/structure_handler.py
```diff
@@ -140,23 +156,25 @@ def execute_one_step(self, configuration: Step,
         # step_executor is [[Step], Step], but referring to Step creates circular dependency
         process = step_executor(configuration)
```
i-keliukh, line 157: I suggest to apply type annotation here

universum/modules/structure_handler.py
```diff
@@ -140,23 +156,25 @@ def execute_one_step(self, configuration: Step,
         # step_executor is [[Step], Step], but referring to Step creates circular dependency
```
i-keliukh, line 156: I think step_executor should be annotated with Callable[[Step], RunningStepBase]

i-keliukh: The comment could be removed in this case

universum/modules/structure_handler.py
```diff
@@ -140,23 +156,25 @@ def execute_one_step(self, configuration: Step,
...
         self.active_background_steps.append({'name': configuration.name,
                                              'block': self.get_current_block(),
-                                             'finalizer': process.finalize,
+                                             'process': process,
                                              'is_critical': configuration.critical})
         return None
 
     def finalize_background_step(self, background_step: BackgroundStepInfo):
-        error = background_step['finalizer']()
+        process = background_step['process']
```
i-keliukh, line 175: Annotation is missing here as well

i-keliukh (approved):

PR #729, opened 2022-08-08
Github Action automation server

dogbert911:
# Description

Automation server for Github Actions

Resolves #(issue)

# How Has This Been Tested?

Start build on Github actions and report build results to review

### DIFF
--- a/universum/configuration_support.py
+++ b/universum/configuration_support.py
@@ -75,6 +75,11 @@ class Step:
         as a separate artifact, so when using ``--no-archive`` option, do not claim directories as `report_artifacts`.
         Please note that any file can be put as `artifact`, `report_artifact`, or both. A file that is both
         in `artifacts` and `report_artifacts`, will be mentioned in a report and will cause build failure when missing.
+
+        .. note::
+
+            GitHub Actions doesn't support this key.
+
     artifact_prebuild_clean
         A flag to signal that artifacts must be cleaned before the build. Cleaning of single files, directories and
         sets of files defined by shell-style patterns is supported. By default, artifacts are not stored in VCS, and
--- a/universum/modules/artifact_collector.py
+++ b/universum/modules/artifact_collector.py
@@ -114,7 +114,7 @@ class ArtifactCollector(ProjectDirectory, HasOutput, HasStructure):
             self.file_list.add(file_name)
             try:
                 file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
-            except RuntimeError as e:
+            except RuntimeError:
                 file_path = file_name
             self.out.log("Adding file " + file_path + " to artifacts...")
             return codecs.open(file_name, "a", encoding="utf-8")          # pylint: disable = consider-using-with
--- a/universum/modules/artifact_collector.py
+++ b/universum/modules/artifact_collector.py
@@ -112,7 +112,10 @@ class ArtifactCollector(ProjectDirectory, HasOutput, HasStructure):
                     raise CriticalCiException(text)
 
             self.file_list.add(file_name)
-            file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
+            try:
+                file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
+            except RuntimeError as e:
+                file_path = file_name
             self.out.log("Adding file " + file_path + " to artifacts...")
             return codecs.open(file_name, "a", encoding="utf-8")          # pylint: disable = consider-using-with
 
--- a/universum/modules/artifact_collector.py
+++ b/universum/modules/artifact_collector.py
@@ -112,10 +112,7 @@ class ArtifactCollector(ProjectDirectory, HasOutput, HasStructure):
                     raise CriticalCiException(text)
 
             self.file_list.add(file_name)
-            try:
-                file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
-            except RuntimeError:
-                file_path = file_name
+            file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
             self.out.log("Adding file " + file_path + " to artifacts...")
             return codecs.open(file_name, "a", encoding="utf-8")          # pylint: disable = consider-using-with
 
--- a/universum/modules/automation_server/github_server.py
+++ b/universum/modules/automation_server/github_server.py
@@ -1,0 +1,49 @@
+from ..error_state import HasErrorState
+from ...lib.module_arguments import ModuleArgumentParser
+from .base_server import BaseServerForHostingBuild, BaseServerForTrigger
+
+__all__ = [
+    "GithubServer",
+]
+
+
+class GithubServer(BaseServerForHostingBuild, BaseServerForTrigger, HasErrorState):
+    @staticmethod
+    def define_arguments(argument_parser: ModuleArgumentParser) -> None:
+        parser = argument_parser.get_or_create_group("Github Actions variables",
+                                                     "Github Actions specific parameters")
+
+        parser.add_argument("--github-server", "-ghs", dest="server_url", metavar="GITHUB_SERVER_URL",
+                            help="Github server URL", type=str)
+        parser.add_argument("--github-repo", "-ghr", dest="repo", metavar="GITHUB_REPOSITORY",
+                            help="Github owner and repository name. For example, octocat/Hello-World", type=str)
+        parser.add_argument("--github-run-id", "-ghri", dest="run_id", metavar="GITHUB_RUN_ID",
+                            help="Github run ID", type=str)
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.check_required_option("server_url", """
+            The URL of the Github server is not specified.
+
+            Please specify the github server url by using '--github-server' ('-ghs') command-line
+            option or GITHUB_SERVER_URL environment variable.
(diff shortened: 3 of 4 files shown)


universum/modules/automation_server/github_server.py
```diff
@@ -0,0 +1,55 @@
+from ..error_state import HasErrorState
+from ...lib.module_arguments import ModuleArgumentParser
+from .base_server import BaseServerForHostingBuild, BaseServerForTrigger
+
+__all__ = [
+    "GithubServer",
+]
+
+
+class GithubServer(BaseServerForHostingBuild, BaseServerForTrigger, HasErrorState):
```
i-keliukh, line 10: You only need to implement BaseServerForTrigger if you are going to support triggering by this server from poller.

universum/modules/artifact_collector.py
```diff
@@ -112,7 +112,10 @@ def create_text_file(self, name):
                     raise CriticalCiException(text)
 
             self.file_list.add(file_name)
-            file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
+            try:
+                file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
+            except RuntimeError:
+                file_path = file_name
```
k-dovgan, line 118: How does this change look when encountered in actual execution?

universum/modules/automation_server/github_server.py
```diff
@@ -0,0 +1,55 @@
...
+    @staticmethod
+    def define_arguments(argument_parser: ModuleArgumentParser) -> None:
+        parser = argument_parser.get_or_create_group("Github Actions variables",
+                                                     "Github Actions specific parameters")
+
+        parser.add_argument("--github-server", "-ghs", dest="server_url", metavar="GITHUB_SERVER_URL",
+                            help="Github server URL", type=str)
+        parser.add_argument("--github-repo", "-ghr", dest="repo", metavar="GITHUB_REPOSITORY",
```
k-dovgan, line 18: Seems I've seen these arguments somewhere else. Are you sure they are not duplicated?

dogbert911: Yes, otherwise we would get a runtime error

k-dovgan: Do you plan to refactor this in separate PR to remove duplication?

dogbert911: No, I don't plan

universum/modules/artifact_collector.py
dogbert911: We cannot provide a link to an individual file in the archive, or to the archive in general, without first creating it (which now looks like unnecessary functionality). therefore, instead of a file reference, we return file path on CI:
```
1.1. Disconnecting
==> Review can be found here: https://swarm.net/reviews/12345/
2. Preparing repository
  ==> Adding file /__w/project/project/artifacts/REPOSITORY_STATE.txt to artifacts...
```

universum/modules/automation_server/github_server.py
k-dovgan: Well, I'd like to leave a note at least somewhere to get back to refactoring it later.

dogbert911: Well, I checked the code, and I'm didn't find same argument. Closest to it is: 

```
("--git-repo", "-gr", dest="repo", metavar="GIT_REPO",
 help="See your project home page for exact repository identifier, passed to 'git clone'. "
      "If using SSH, '--git-repo' format is 'ssh://user@server:port/detailed/path'")
```

But it's related to GIT VCS not for CI.  
So I think we don't need to mention this "duplication". Because it's possible to run universum on GA (from one repo) and download  sources from another.

k-dovgan: Well, then I'd think more about totally different parameters 'git repo' and 'github repo' being called so similarly. I don't have ideas at the moment, but I can clearly see how it can become a problem for the user.

dogbert911: What if I just renames parameters to `--github-actions-server|-gas`, `--github-actions-repo|-gar` and etc?

i-keliukh: In my opinion it should be --github-repo. However, we can add notes to distinguish this and git repo. Note should be added to both parameters.

I prefer shorter form because in the future it can be used for some other case.

universum/modules/automation_server/github_server.py
```diff
@@ -0,0 +1,58 @@
...
+]
+
+
+class GithubServer(BaseServerForHostingBuild, BaseServerForTrigger, HasErrorState):
+    @staticmethod
+    def define_arguments(argument_parser: ModuleArgumentParser) -> None:
+        parser = argument_parser.get_or_create_group("Github Actions variables",
+                                                     "Github Actions specific parameters")
+
+        parser.add_argument("--github-server", "-ghs", dest="server_url", metavar="GITHUB_SERVER_URL",
```
i-keliukh, line 16: I propose to set the default value to https://github.com

universum/modules/automation_server/github_server.py
```diff
@@ -0,0 +1,58 @@
...
+                            help="Github server URL, used on Github Actions for constructing links to builds. "
+                                 "For example: https://github.com", type=str)
+        parser.add_argument("--github-repo", "-ghr", dest="repo", metavar="GITHUB_REPOSITORY",
+                            help="Github owner and repository name, used on Github Actions for constructing links to "
+                                 "builds. For example: octocat/Hello-World", type=str)
+        parser.add_argument("--github-run-id", "-ghri", dest="run_id", metavar="GITHUB_RUN_ID",
```
i-keliukh, line 22: For me it is kind of hard to understand this parameter without some page that describes using of universum in GitHub Actions. Is it supposed to be obtained from some GA parameters during launch? Which one? What is the recommended command line then?

universum/modules/automation_server/github_server.py
dogbert911: We don't need this, because we use one of GA default variables:
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

i-keliukh: Also I propose to stop using multi-character parameter names for compact form (like -ghs, -ghr, -ghri).
Traditionally linux utilities use single-character compact form case sensitive flags.
We already violate this rule in multiple places, but at least we can stop making things worse.

i-keliukh: Our options are:
1. Use some semi-random case-sensitive letters for compact form flags
2. Drop compact form

i-keliukh: > We don't need this, because we use one of GA default variables:
Well, ok. Then we need to correct help message to mention that the GA sets this variable already and typically user doesn't need to set any of these parameters.

universum/modules/automation_server/github_server.py
dogbert911: All this parameters are getting from GA environment variables, and their are not recommended  to be set on the command line.

universum/modules/automation_server/github_server.py
dogbert911: > Also I propose to stop using multi-character parameter names for compact form (like -ghs, -ghr, -ghri). Traditionally linux utilities use single-character compact form case sensitive flags. We already violate this rule in multiple places, but at least we can stop making things worse.

Ok, I was just looking at other arguments

> Well, ok. Then we need to correct help message to mention that the GA sets this variable already and typically user doesn't need to set any of these parameters.

Ok

universum/modules/artifact_collector.py
```diff
@@ -112,7 +112,10 @@ def create_text_file(self, name):
                     raise CriticalCiException(text)
 
             self.file_list.add(file_name)
-            file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
+            try:
+                file_path = self.automation_server.artifact_path(self.artifact_dir, os.path.basename(file_name))
```
i-keliukh, line 116: It seems there are two issues here:
1. Printing any path doesn't make much sense for GitHub Actions at the moment
2. The logic is based on exceptions
We cannot do much about item 1. But we can just return the name from the GithubServer class instead of raising an exception. The end result will be the same (or almost the same if you cannot recover the original path inside GithubServer).

universum/configuration_support.py
```diff
@@ -75,6 +75,11 @@ class Step:
         as a separate artifact, so when using ``--no-archive`` option, do not claim directories as `report_artifacts`.
         Please note that any file can be put as `artifact`, `report_artifact`, or both. A file that is both
         in `artifacts` and `report_artifacts`, will be mentioned in a report and will cause build failure when missing.
+
+        .. note::
+
+            GitHub Actions doesn't support this key.
```
i-keliukh, line 81: As far as I understand the artifacts key also is not supported by GitHub Actions.
Is that correct?

dogbert911: It is not correct. The `artifacts` key copies files to the artifacts folder, and GA can store artifacts from there.

i-keliukh: Just a note to myself: regular artifacts are not reported in the build log and for them the GitHub Action implementation is not called.

i-keliukh:
I believe the documentation update is a must for this change

dogbert911:
> I believe the documentation update is a must for this change

I believe too :), but in another PR (still in progress)

dogbert911:
GHA documentation PR: vhttps://github.com/Samsung/Universum/pull/734

i-keliukh (approved):

PR #731, opened 2022-08-19
fix(analyzer): wrong error on missing input file

i-keliukh:
When the passed input file is missing, there were no explicit error
message. Instead, the empty list of files to analyze was passed to the
actual analysis tool. As the result, the error message was incorrect and
hard to understand. Fixed by producing warning on each missing input
file and producing error if the final list is empty.

### DIFF
--- a/universum/analyzers/utils.py
+++ b/universum/analyzers/utils.py
@@ -110,7 +110,15 @@ def expand_files_argument(settings: argparse.Namespace) -> None:
     # TODO: subclass argparse.Action
     result = []
     for pattern in settings.file_list:
-        result.extend(glob.glob(pattern))
+        file_list: List[str] = glob.glob(pattern)
+        if not file_list:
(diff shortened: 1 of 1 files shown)


k-dovgan (dismissed):

universum/analyzers/utils.py
```diff
@@ -110,7 +110,15 @@ def expand_files_argument(settings: argparse.Namespace) -> None:
     # TODO: subclass argparse.Action
     result = []
```
o-andrieiev, line 111: why not mark result as List[str]?

i-keliukh: BTW, shouldn't it be set instead of list?

o-andrieiev: Indeed, set seems like a better fit

i-keliukh: Fixed

k-dovgan (approved):

PR #732, opened 2022-08-22
refactor(output): rename and reorder methods

i-keliukh:
1. Reordered and grouped methods, renamed them to make naming consistent. Now names correspond to what methods really do.
2. Added type annotations, documentation and fixed issues found by PyCharm inspections for affected files
3. Made some methods protected
4. Changed interface of the "log_summary_step" (previously report_step) function to make it explicit

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -13,6 +13,7 @@ Project 'Universum'
     code_report.rst
     github_handler.rst
     universum_docs.rst
+    internal.rst
     teamcity.rst
     jenkins.rst
     examples.rst
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -13,10 +13,10 @@ Project 'Universum'
     code_report.rst
     github_handler.rst
     universum_docs.rst
-    internal.rst
     teamcity.rst
     jenkins.rst
     examples.rst
+    internal.rst
     changelog_ref.rst
 
 .. image:: _static/logo.svg
--- a/universum/modules/code_report_collector.py
+++ b/universum/modules/code_report_collector.py
@@ -5,14 +5,14 @@ import urllib.parse
 from copy import deepcopy
 from typing import Dict, List, Optional, TextIO, Tuple
 
-from ..configuration_support import Configuration
+from . import artifact_collector, reporter
 from .output import HasOutput
 from .project_directory import ProjectDirectory
-from . import artifact_collector, reporter
+from .structure_handler import HasStructure
+from ..configuration_support import Configuration
 from ..lib import utils
 from ..lib.gravity import Dependency
 from ..lib.utils import make_block
-from .structure_handler import HasStructure
 
 
 class CodeReportCollector(ProjectDirectory, HasOutput, HasStructure):
@@ -95,7 +95,7 @@ class CodeReportCollector(ProjectDirectory, HasOutput, HasStructure):
             json_file: TextIO = self.artifacts.create_text_file("Static_analysis_report.json")
             json_file.write(json.dumps(report, indent=4))
 
-            issue_count: int = 0
+            issue_count: int
             if report or report == []:
                 try:
                     issue_count = self._report_as_sarif_json(report)
--- a/universum/modules/code_report_collector.py
+++ b/universum/modules/code_report_collector.py
@@ -103,15 +103,15 @@ class CodeReportCollector(ProjectDirectory, HasOutput, HasStructure):
                     try:
                         issue_count = self._report_as_pylint_json(report)
                     except (KeyError, AttributeError, ValueError):
-                        self.out.log_stderr("Could not parse report file. Something went wrong.")
+                        self.out.log_exception("Could not parse report file. Something went wrong.")
                         continue
             else:
-                self.out.log_stderr("There are no results in code report file. Something went wrong.")
+                self.out.log_exception("There are no results in code report file. Something went wrong.")
                 continue
 
             if issue_count != 0:
                 text = str(issue_count) + " issues"
-                self.out.log_stderr("Found " + text)
+                self.out.log_exception("Found " + text)
                 self.out.report_build_status(os.path.splitext(os.path.basename(report_file))[0] + ": " + text)
             else:
                 self.out.log("Issues not found.")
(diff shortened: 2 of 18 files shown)


universum/modules/output/base_output.py
```diff
@@ -3,46 +3,130 @@
...
+        :param step_title: the title of the step
+        :param has_children: whether the step has children or not. Typically, if step has children, the status is not
+            included in the summary.
+        :param status: the status of the step: "Passed", "Failed" or "Skipped"
+        """
         raise NotImplementedError
 
     def report_build_problem(self, description: str) -> None:
+        """
+        TeamCity only. Report run problem to the TeamCity via the output. Also fails the build regardless of the process
```
miltolstoy, line 115: Hmm, maybe rename this method to `report_teamcity_build_problem`? The information that this and the next methods are TeamCity-only are quite hidden.

miltolstoy (commented):
LGTM, but left some thoughts in the comments.

universum/modules/output/base_output.py
i-keliukh: I thought about it, but decided not to do it. Reasons are the following:
1. At some point we may support it for more Jenkins and other servers
2. If it is only for TeamCity it should not be part of the common interface at all. But at the moment it doesn't make sense to separate just two methods.

doc/index.rst
```diff
@@ -13,6 +13,7 @@ Project 'Universum'
     code_report.rst
     github_handler.rst
     universum_docs.rst
+    internal.rst
```
k-dovgan, line 16: I'd actually moved this to before examples. This is not the most interesting docs for user: interaction with CodeReview systems seems more important to me.

universum/modules/code_report_collector.py
```diff
@@ -95,23 +95,23 @@ def report_code_report_results(self) -> None:
             json_file: TextIO = self.artifacts.create_text_file("Static_analysis_report.json")
             json_file.write(json.dumps(report, indent=4))
 
-            issue_count: int = 0
+            issue_count: int
             if report or report == []:
```
o-andrieiev, line 99: how about inverting this condition to reduce indentation?

universum/modules/output/base_output.py
```diff
@@ -3,52 +3,138 @@
...
+        """
+        Print one line of the standard output of the previously launched external command.
+
+        :param line: the line to print
+        """
         raise NotImplementedError
 
-    def report_step(self, message, status):
-        self.log(message)
+    def log_stderr(self, line: str) -> None:
```
o-andrieiev, line 63: I find that log_stderr and log_error lack distinction, could we make use of each a bit clearer?

o-andrieiev: same with log and log_stdout

universum/modules/output/html_output.py
```diff
@@ -1,106 +1,103 @@
...
-
-    def open_block(self, num_str, name):
-        opening_html = f'<input type="checkbox" id="{num_str}" class="hide"/>' + \
-            f'<label for="{num_str}"><span class="sectionLbl">'
-        closing_html = "</span></label><div>"
-        name_html = f'<span class="sectionTitle">{name}</span>'
-        self._log_line(f"{opening_html}{num_str} {name_html}{closing_html}", with_line_separator=False)
-        self._block_level += 1
+        self._log_name: str = log_name
+        self._log_path: Optional[str] = None
```
o-andrieiev, line 22: not Optional[Path]?

universum/modules/output/html_output.py
```diff
@@ -110,47 +107,49 @@ def _log_buffered(self, line):
...
+
         with open(self._log_path, "a", encoding="utf-8") as file:
             file.write(line)
 
-    def _build_indent(self):
-        indent_str = []
-        for x in range(0, self._block_level):
+    def _build_indent(self) -> str:
+        indent_str: List[str] = []
+        for x in range(0, self._block_level):  # type: int
```
o-andrieiev, line 124: do you really need to specify type here? I'd assume that mypy figures it out based on range() return

universum/modules/output/terminal_based_output.py
```diff
@@ -19,78 +20,78 @@ class Colors:
...
     @staticmethod
-    def stdout(*args, **kwargs):
+    def _stdout(*args, **kwargs) -> None:
         sys.stdout.write(''.join(args))
         if not kwargs.get("no_enter", False):
             sys.stdout.write('\n')
 
-    def indent(self):
+    def _indent(self) -> None:
         for x in range(0, self.block_level):
```
o-andrieiev, line 35: you added annotation in similar case html_output.py#124

universum/modules/code_report_collector.py
```diff
@@ -95,23 +95,23 @@ def report_code_report_results(self) -> None:
...
-                self.out.log_stderr("There are no results in code report file. Something went wrong.")
+                self.out.log_error("There are no results in code report file. Something went wrong.")
                 continue
 
             if issue_count != 0:
                 text = str(issue_count) + " issues"
-                self.out.log_stderr("Found " + text)
-                self.out.report_build_status(os.path.splitext(os.path.basename(report_file))[0] + ": " + text)
+                self.out.log_error("Found " + text)
+                self.out.set_build_title(os.path.splitext(os.path.basename(report_file))[0] + ": " + text)
```
k-dovgan, line 115: "Build title" can be mistaken for build tag. I guess this thing might be called 'status line' or something like that. Maybe, 'inline report'?

i-keliukh: It was previously called "build status" and I deliberately changed it.
This is because it is literally a title of this build on the build history page in the TeamCity.

universum/modules/output/base_output.py
i-keliukh: Isn't the doc string enough?
The idea is that log_stdout and log_stderr used for printing respective output of external commands.
log and log_error are for producing own messages.

universum/modules/code_report_collector.py
k-dovgan: TeamCity says just 'Status', so we can use the official naming.

universum/modules/output/html_output.py
i-keliukh: No. It is produced by os.path.join, which returns str.
Switching to Path or py is a different kind of refactoring.

k-dovgan (approved):

miltolstoy (approved):

o-andrieiev (approved):
LGTM

PR #736, opened 2022-09-23
Fails only short report

dogbert911:
# Description

Add options to print falis_only short report (print only real failed steps)

Resolves #(issue) None

### DIFF
--- a/universum/modules/reporter.py
+++ b/universum/modules/reporter.py
@@ -156,7 +156,7 @@ class Reporter(HasOutput, HasStructure):
             text += indent + str(block) + '\n'
             self.out.log_summary_step(indent + block_title, has_children, block.status)
         elif not block.is_successful():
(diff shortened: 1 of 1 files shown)


universum/modules/reporter.py
```diff
@@ -151,8 +156,9 @@ def _report_steps_recursively(self, block: Block, text: str, indent: str) -> Tup
             text += indent + str(block) + '\n'
             self.out.log_summary_step(indent + block_title, has_children, block.status)
         elif not block.is_successful():
-            text += str(block) + '\n'
-            self.out.log_summary_step(block_title, has_children, block.status)
+            if not self.settings.only_fails_short or not block.children:
```
i-keliukh, line 159: There is already `has_children` variable defined within this function

dogbert911: tnx

i-keliukh (approved):

PR #737, opened 2022-10-05
feat(structure_handler): initialize conditional steps logic

miltolstoy:
One more merge of [this](https://github.com/Samsung/Universum/pull/710) PR changes.
Relates to https://github.com/Samsung/Universum/issues/709

### DIFF
--- a/universum/configuration_support.py
+++ b/universum/configuration_support.py
@@ -170,6 +170,8 @@ class Step:
                  pass_tag: str = '',
                  fail_tag: str = '',
                  if_env_set: str = '',
+                 if_succeeded = None,
+                 if_failed = None,
                  **kwargs) -> None:
         self.name: str = name
         self.directory: str = directory
@@ -185,6 +187,9 @@ class Step:
         self.pass_tag: str = pass_tag
         self.fail_tag: str = fail_tag
         self.if_env_set: str = if_env_set
+        self.if_succeeded = if_succeeded
+        self.if_failed = if_failed
+        self.is_conditional = self.if_succeeded or self.if_failed
         self.children: Optional['Configuration'] = None
         self._extras: Dict[str, str] = {}
         for key, value in kwargs.items():
@@ -391,6 +396,9 @@ class Step:
             pass_tag=self.pass_tag + other.pass_tag,
             fail_tag=self.fail_tag + other.fail_tag,
             if_env_set=self.if_env_set + other.if_env_set,
+            # FIXME: This is a dummy implementation. Define addition logic and implement it.
+            if_succeeded=other.if_succeeded,
+            if_failed=other.if_failed,
             **combine(self._extras, other._extras)
         )
 
--- a/universum/configuration_support.py
+++ b/universum/configuration_support.py
@@ -115,9 +115,9 @@ class Step:
         A tag used to mark successful TeamCity builds. This tag can be set independenty
         of `fail_tag` value per each step. The value should be set to a strings without spaces as acceptable by
         TeamCity as tags. Every tag is added (if matching condition) after executing build step it is set in,
-        not in the end of all run.
+        not in the end of all run. Not applicable for conditional steps.
     fail_tag
-        A tag used to mark failed TemCity builds. See `pass_tag` for details.
+        A tag used to mark failed TemCity builds. See `pass_tag` for details. Not applicable for conditional steps.
 
     Each parameter is optional, and is substituted with a falsy value, if omitted.
 
--- a/universum/modules/launcher.py
+++ b/universum/modules/launcher.py
@@ -248,16 +248,6 @@ class RunningStep(RunningStepBase):
         else:
             self.out.log_stderr(line)
 
-    def add_tag(self, tag: str) -> None:
-        if not tag:
-            return
-
-        request: Response = self.send_tag(tag)
-        if request.status_code != 200:
-            self.out.log_error(request.text)
-        else:
-            self.out.log("Tag '" + tag + "' added to build.")
-
     def finalize(self) -> None:
         self._error = None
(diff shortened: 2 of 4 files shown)


universum/configuration_support.py
```diff
@@ -391,6 +396,9 @@ def __add__(self, other: 'Step') -> 'Step':
             pass_tag=self.pass_tag + other.pass_tag,
             fail_tag=self.fail_tag + other.fail_tag,
             if_env_set=self.if_env_set + other.if_env_set,
+            # FIXME: This is a dummy implementation. Define addition logic and implement it.
```
i-keliukh, line 399: Please add the description of possible variants of combining conditional steps with groups/child steps to some task.

universum/modules/launcher.py
```diff
@@ -282,8 +282,9 @@ def finalize(self) -> None:
                 text = utils.trim_and_convert_to_unicode(text)
                 if self.file:
                     self.file.write(text + "\n")
-                self.add_tag(self.configuration.fail_tag)
-                self._error = text
+                if not self.configuration.is_conditional:
```
i-keliukh, line 285: Is the logic that since the conditional step always succeeds (kind of), there is no sense to use tagging for it?

i-keliukh: Ah, nope. The "pass_tag" is supported though.
Looks like inconsistency.

universum/modules/structure_handler.py
```diff
@@ -264,6 +265,20 @@ def execute_steps_recursively(self, parent: Step,
...
+    def execute_conditional_step(self, step, step_executor):
+        self.configs_current_number += 1
+        step_name = self._build_step_name(step.name)
+        conditional_step_succeeded = False
+        with self.block(block_name=step_name, pass_errors=True):
+            process = self.execute_one_step(step, step_executor);
+            conditional_step_succeeded = not process.get_error()
+        step_to_execute = step.if_succeeded if conditional_step_succeeded else step.if_failed
+        return self.execute_steps_recursively(parent=Step(),
+                                              children=Configuration([step_to_execute]),
```
i-keliukh, line 278: This assumes that only single step is supported in if_succeeded/if_failed parameters.
I propose to support whole Configuration instead.

universum/modules/structure_handler.py
```diff
@@ -282,7 +297,10 @@ def report_background_steps(self) -> bool:
         return result
 
     def execute_step_structure(self, configs: Configuration, step_executor) -> None:
-        self.configs_total_count = sum(1 for _ in configs.all())
+        for config in configs.all():
+            self.configs_total_count += 1
+            if config.if_succeeded or config.if_failed:
```
i-keliukh, line 302: Can you use config.is_conditional here instead?

universum/configuration_support.py
miltolstoy: Multiple high-level work items (alongside with the task you mentioned) were recorded in the corresponding [task](https://github.com/Samsung/Universum/issues/709). I would like to investigate possible combining variants later when I will start with that work item.

universum/modules/structure_handler.py
miltolstoy: The "implement Configuration support instead of Step for branch steps" entry was added to the [task](https://github.com/Samsung/Universum/issues/709).

universum/modules/structure_handler.py
```diff
@@ -264,6 +265,20 @@ def execute_steps_recursively(self, parent: Step,
...
 
+
+    def execute_conditional_step(self, step, step_executor):
+        self.configs_current_number += 1
+        step_name = self._build_step_name(step.name)
+        conditional_step_succeeded = False
+        with self.block(block_name=step_name, pass_errors=True):
+            process = self.execute_one_step(step, step_executor);
+            conditional_step_succeeded = not process.get_error()
+        step_to_execute = step.if_succeeded if conditional_step_succeeded else step.if_failed
```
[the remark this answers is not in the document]
miltolstoy: It will be handled as a separate work item ("check work with only one branch step specified") under the general [task](https://github.com/Samsung/Universum/issues/709).

universum/modules/launcher.py
```diff
@@ -308,6 +295,19 @@ def _handle_postponed_out(self) -> None:
             item[0](item[1])
         self._postponed_out = []
 
+    def _get_teamcity_build_tag(self) -> Optional[str]:
+        if self.configuration.is_conditional:
+            return None  # conditional steps always succeed, no sense to set a tag
+        tag: str = self.configuration.fail_tag if self._error else self.configuration.pass_tag
+        return tag  # can be also None if not set for current Configuration
+
+    def _assign_teamcity_build_tag(self, tag: str) -> None:
```
miltolstoy, line 304: Used wording from the TeamCity website: https://blog.jetbrains.com/teamcity/2010/03/tags-in-teamcity/

universum/modules/launcher.py
```diff
@@ -282,14 +272,12 @@ def finalize(self) -> None:
...
                     self.file.write(text + "\n")
-                self.add_tag(self.configuration.fail_tag)
                 self._error = text
-                return
-
-            self.add_tag(self.configuration.pass_tag)
-            return
 
         finally:
+            tag: Optional[str] = self._get_teamcity_build_tag()
```
i-keliukh, line 278: For me the benefit of having two functions and storing the result in a variable is not clear

i-keliukh: Ok, just for reference: the returned tag can be None even without additional check of condition, just because the tag is not set in the configuration. This makes the introduction of variable more reasonable.

i-keliukh (approved):

k-dovgan (approved):

PR #745, opened 2022-11-10
feat(report): return non-zero exit code if any step fails

### LINKED ISSUE #743: Ability to return non-success exit code if some build steps are failed
**Describe the desirable behavior of the program**

There is no ability to inform the parent process (that launches Universum execution) about failed steps inside. 
Please add some command line option that enables build steps status-related exit code.

**Additional context**

When using Jenkins and the Stage View, it is not possible to see that some stage, where Universum is used, was completed with build step errors. Stages are always green and successful

k-dovgan:
Resolves #743

### DIFF
--- a/universum/modules/reporter.py
+++ b/universum/modules/reporter.py
@@ -6,6 +6,7 @@ from typing_extensions import TypedDict
 from . import automation_server
 from .output import HasOutput
 from .structure_handler import HasStructure, Block
+from ..lib.ci_exception import SilentAbortException
 from ..lib.gravity import Dependency
 from ..lib.utils import make_block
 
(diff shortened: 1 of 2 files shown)


miltolstoy:
LGTM

i-keliukh (approved):
Approving because it is urgent

universum/modules/reporter.py
```diff
@@ -149,6 +154,9 @@ def report_build_result(self):
             for observer in self.observers:
                 observer.code_report_to_review(self.code_report_comments)
 
+        if self.settings.fail_unsuccessful and not is_successful:
```
i-keliukh, line 157: Duplication of code is clear indication that something is wrong with this approach

universum/modules/reporter.py
```diff
@@ -119,6 +122,8 @@ def report_build_result(self):
 
         if not self.observers:
             self.out.log("Nowhere to report. Skipping...")
+            if self.settings.fail_unsuccessful and not is_successful:
```
i-keliukh, line 125: I think that setting of exit code has to be function of main.py, not reporter.
main.pu can use data from reporter for this, but it has to set the exit code on his own.

i-keliukh: Otherwise, you will not be able to change something in main.py.
For example, you cannot add any code after reporter, because it will not be run due to exception.
I propose to commit this as is, and refactor in the next PR.

PR #751, opened 2022-12-24
add(github): enable actions workflow

k-dovgan:


.github/workflows/precommit-check.yml
```diff
@@ -0,0 +1,38 @@
+name: Universum check
+on: push
```
dogbert911, line 2: You use trigger on push, it works for any push in any branches.
Probably better to use next triggers - PR to master, and push to master:
```
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
```

k-dovgan: Actually, I chose 'on: push' on purpose. This is the behavior we had before Actions: any commit to repository was checked.
And, as any commit is checked, no further checks for pull requests are required.
The only thing I don't like is that if I push more than one commit at a time, only the last one will be checked. I'd like every single commit to be checked.

axxie: IMO it is nice feature for users. If they want every commit checked they can push them separately. If they are only interested in the latest commit status, as they should, they can push them together.

i-keliukh: IMO it is nice feature for users. If they want every commit checked they can push them separately. If they are only interested in the latest commit status, as they should, they can push them together.

.github/workflows/precommit-check.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Setup python 3.8
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
+
+    - name: Install dependency
+      run: pip install universum[test]
```
i-keliukh, line 16: Are you sure it uses the version from the repository?

i-keliukh (approved):

PR #755, opened 2023-01-19
fix(checks): MyPy and Pylint warnings

k-dovgan:
Also turn on failing on failed steps for better GhA reporting

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -109,9 +109,9 @@ configs = Configuration([dict(name="Test configuration", command=["ls", "-la"])]
 
 
 class BaseVcsClient:
-    def __init__(self):
-        self.root_directory: py.path.local  # type: ignore
-        self.repo_file: py.path.local  # type: ignore
+    def __init__(self) -> None:
+        self.root_directory: py.path.local
+        self.repo_file: py.path.local
 
     def get_last_change(self):
         pass
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -116,13 +116,13 @@ class BaseVcsClient:
     def get_last_change(self):
         pass
 
-    def file_present(self, file_path: str) -> bool:
+    def file_present(self, file_path: str) -> bool:  # type: ignore
         pass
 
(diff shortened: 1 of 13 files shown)


miltolstoy (commented):
LGTM

i-keliukh (changes requested):
Please fix mypy warnings as well

tests/utils.py
```diff
@@ -110,8 +110,8 @@ def create_empty_settings(test_type: str) -> ModuleNamespace:
 
 class BaseVcsClient:
     def __init__(self):
-        self.root_directory: py.path.local
-        self.repo_file: py.path.local
+        self.root_directory: py.path.local  # type: ignore
```
i-keliukh, line 113: In my opinion it would be much better to annotate the function than to suppress warning. The warning is valid.

universum/modules/reporter.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, *args, **kwargs):
         self.report_initialized = False
         self.blocks_to_report = []
         self.artifacts_to_report = []
-        self.code_report_comments: Dict[str, List[ReportMessage]] = defaultdict(list)
+        self.code_report_comments: Dict[str, List[ReportMessage]] = defaultdict(list)  # type: ignore
```
i-keliukh, line 66: Same issue: it is better to annotate function than to ignore warning

universum/modules/vcs/perforce_vcs.py
```diff
@@ -324,8 +324,8 @@ def __init__(self, *args, **kwargs):
         self.client_view = []
         self.mappings_dict = {}
 
-        self.unshelved_files: List[Dict[str, str]] = []
-        self.diff_in_files: List[Tuple[Optional[str], Optional[str], Optional[str]]] = []
+        self.unshelved_files: List[Dict[str, str]] = []  # type: ignore
```
i-keliukh, line 327: Same issue

i-keliukh (approved):

PR #756, opened 2023-01-25
test(artifact_collector): preprocessing artifacts in the 'artifacts' directory

miltolstoy:
+ wrong typing fixed

### DIFF
--- a/tests/test_preprocess_artifacts.py
+++ b/tests/test_preprocess_artifacts.py
@@ -2,8 +2,8 @@
 
 import inspect
 import os
-import pathlib
 from typing import Generator
+from py._path.local import LocalPath
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


tests/test_preprocess_artifacts.py
```diff
@@ -2,8 +2,8 @@
 
 import inspect
 import os
-import pathlib
 from typing import Generator
+from py._path.local import LocalPath
```
k-dovgan, line 6: Can you please comment this decision? Is this because tmpdir creates py._path.local.LocalPath objects, or for some other reason?

tests/test_preprocess_artifacts.py
```diff
@@ -62,17 +63,33 @@ def test_no_artifact_no_prebuild_clean(test_env: ArtifactsTestEnvironment) -> No
...
+
+def test_artifact_in_artifacts_dir_prebuild_clean(test_env: ArtifactsTestEnvironment,
+                                                  stdout_checker: FuzzyCallChecker) -> None:
+    test_env.write_config_file(artifact_prebuild_clean=True)
+    test_env.create_artifact_file(test_env.artifact_dir)
+    test_env.run(expect_failure=True)
+    stdout_checker.assert_has_calls_with_param("already present in artifact directory")
+
+
+def test_artifact_in_artifacts_dir_no_prebuild_clean(test_env: ArtifactsTestEnvironment,
```
i-keliukh, line 90: If the two adjacent tests are almost identical, doesn't it make sense to use parametrization?

i-keliukh (approved):

PR #757, opened 2023-01-26
test: use `tmp_path` fixture instead of `tmpdir` 

miltolstoy:


tests/deployment_utils.py
```diff
@@ -154,26 +155,27 @@ def clean_execution_environment(request):
...
         except OSError:
             pass
-        source_dir.ensure(dir=True)
+        source_dir.mkdir()
 
     else:
-        source_dir = tmpdir.mkdir("project_sources")
-    local_file = source_dir.join("readme.txt")
-    local_file.write("This is a an empty file")
+        source_dir = tmp_path.joinpath("project_sources")
```
i-keliukh, line 175: Wouldn't it be better to write `tmp_path / "project_sources"` instead?

tests/deployment_utils.py
```diff
@@ -165,7 +166,7 @@ def local_sources(tmp_path: py.path.local):
     if utils.reuse_docker_containers():
         source_dir = py.path.local(".work")
         try:
-            source_dir.remove(rec=1, ignore_errors=True)
+            shutil.rmtree(str(source_dir), ignore_errors=True)
```
i-keliukh, line 169: Cast to str is not needed

tests/test_p4_exception_handling.py
```diff
@@ -27,7 +27,7 @@ def test_p4_forbidden_local_revert(perforce_environment: P4TestEnvironment, stdo
     perforce_environment.shelve_config(config)
     result = __main__.run(perforce_environment.settings)
     # Clean up the directory at once to make sure it doesn't remain non-writable even if some assert fails
-    perforce_environment.temp_dir.chmod(0o0777, rec=1)
+    perforce_environment.temp_dir.chmod(0o0777)
```
i-keliukh: It was recursive in previous version. Not sure whether it is critical or not.

miltolstoy: Was fixed in [one](https://github.com/Samsung/Universum/pull/757/commits/0ed013fff73236a720a7028f6958f4cf732f4fb4) of the following commits

tests/deployment_utils.py
```diff
@@ -2,10 +2,11 @@
 
 import getpass
 import os
+import shutil
+import pathlib
```
k-dovgan, line 6: We have several places in repo where `from pathlib import Path` is used. Maybe we want to standardize Path usage in this very commit?

tests/test_check_if_env_set.py
```diff
@@ -1,7 +1,7 @@
 # pylint: disable = redefined-outer-name
 
 import os
-from typing import Optional
+from typing import Optional, List
```
k-dovgan, line 4: Why did it work before this change?

i-keliukh (approved):

PR #758, opened 2023-02-02
fix(checks): ignore Pylint warnings for better readability

k-dovgan:


universum/modules/output/terminal_based_output.py
```diff
@@ -23,7 +23,7 @@ class TerminalBasedOutput(BaseOutput):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self.block_level = 0
-        self.unicode_acceptable = (locale.getpreferredencoding() == "UTF-8")
+        self.unicode_acceptable = (locale.getpreferredencoding() == "UTF-8")  # pylint: disable = superfluous-parens
```
i-keliukh, line 26: Why this warning is not correct?

k-dovgan: It is totally correct, it's just expression is more readable with parentheses

miltolstoy (commented):
LGTM

i-keliukh (approved):

PR #759, opened 2023-02-03
test(artifact_collector): preprocessing directory artifacts

miltolstoy:


tests/test_preprocess_artifacts.py
```diff
@@ -18,18 +19,36 @@ def __init__(self, tmp_path: pathlib.Path) -> None:
...
     def write_config_file(self, artifact_prebuild_clean: bool) -> None:
+        artifact_in_dir = f"{self.dir_name}/{self.artifact_name}"
         config: str = inspect.cleandoc(f"""
             from universum.configuration_support import Configuration, Step
-            step = Step(name='Step',
-                        command=['bash', '-c', 'echo "{self.artifact_content}" > {self.artifact_name}'],
-                        artifacts='{self.artifact_name}',
-                        artifact_prebuild_clean={artifact_prebuild_clean})
-            configs = Configuration([step])
+            step_with_file = Step(name='Step with file',
```
k-dovgan, line 29: I would use the same formatting for step_with_file and step_with_dir claming; but that's a subjective preference.

k-dovgan (dismissed):

tests/test_preprocess_artifacts.py
```diff
@@ -39,10 +58,16 @@ def check_step_artifact_present(self) -> None:
     def check_step_artifact_absent(self) -> None:
         assert not os.path.exists(self.artifact_path)
 
-    def create_artifact_file(self, directory: pathlib.Path) -> None:
-        precreated_artifact: pathlib.Path = directory / self.artifact_name
-        with open(precreated_artifact, "w", encoding="utf-8") as f:
-            f.write("pre-created artifact content")
+    def check_step_dir_artifact_present(self) -> None:
+        assert os.path.exists(self.dir_archive)
+        with zipfile.ZipFile(self.dir_archive) as dir_zip:
```
i-keliukh, line 63: This check assumes '--no-archive' option is absent by default.
This could change in the future.
I suggest to explicitly set it in settings to False or to implement tests for both cases.

i-keliukh (approved):

PR #760, opened 2023-02-07
fix(report): set exit code even if reporting crashes

k-dovgan:
1. Move 'fail_unsuccessful' option that changes program behavior to Main module
2. Change 'pass_errors=False' in result reporting to explicit exception handling for the guaranteed return value
3. Fix bug caused by p.2
4. Fix typing warning

### DIFF
--- a/universum/main.py
+++ b/universum/main.py
@@ -43,6 +43,8 @@ class Main(HasOutput):
                                      help="Only applies to build steps where ``code_report=True``; "
                                           "disables calculating analysis diff for changed files, "
                                           "in this case full analysis report will be published")
+        argument_parser.add_argument("--fail-unsuccessful", action="store_true", dest="fail_unsuccessful",
+                                     help="Return non-zero exit code if any step failed")
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


universum/modules/reporter.py
```diff
@@ -154,8 +152,9 @@ def report_build_result(self):
             for observer in self.observers:
                 observer.code_report_to_review(self.code_report_comments)
 
-        if self.settings.fail_unsuccessful and not is_successful:
-            raise SilentAbortException(1)
+        if is_successful:
```
i-keliukh, line 155: This block could be simplified to "return is_successful"

universum/main.py
```diff
@@ -88,7 +90,9 @@ def execute(self) -> None:
                     self.code_report_collector.repo_diff = repo_diff
             self.code_report_collector.report_code_report_results()
         self.artifacts.report_artifacts()
-        self.reporter.report_build_result()
+        result = self.reporter.report_build_result()
+        if self.settings.fail_unsuccessful and not result:
```
i-keliukh, line 94: Does this condition rely on "block" context manager with pass_error=False returning None (not returning anything) in case of exception?
If that is the case, I would say this implicit dependency is far from obvious.

i-keliukh:
If the change is fix, it would be better to include the description of what is fixed into commit message, like below:
"fix(report): set exit code even if reporting crashes"

i-keliukh:
What about nonci?
Previous implementation tried to set exit code for nonci too.

k-dovgan:
> What about nonci? Previous implementation tried to set exit code for nonci too.

That was a side effect, not really planned to implement right now.

i-keliukh (approved):

PR #766, opened 2023-02-13
test(artifact_collector): .zip file in artifacts directory

miltolstoy:


k-dovgan (approved):

miltolstoy:
This PR shows a bug in the `ArtifactCollector.preprocess_artifact_list()` - a .zip file presence is checked not only for a directory artifact but for a file artifact also. Will be fixed in another PR.

Issue #770, opened 2023-03-15
Uncrustify usability issues


i-keliukh:
The uncrustify analyzer doesn't deal well with error conditions.
The issues are described below.

# File path handling

## Steps to reproduce

1. Create some poorly formatted `test.c` file 
2. Create empty `uncrustify.cfg` file
3. Install uncrustify (`sudo apt install uncrustify`)
4. Launch the uncrustify analyzer like follows:

```shell
$ python -m universum.analyzers.uncrustify --files $(pwd)/test.c --result-file res.json --cfg-file $(pwd)/uncrustify.cfg
```

## Expected result

The analyzer finishes successfully or produces actionable error message.

## Actual result

The analyzer produces the following error message:

```
[Errno 2] No such file or directory: '/home/user/Projects/common/format-test/uncrustify/test.c'
```

## Details

The issue happens because `--files` argument contains the full path. If using relative path, all works correctly:

```shell
python -m universum.analyzers.uncrustify --files test.c --result-file res.json --cfg-file $(pwd)/uncrustify.cfg
<no error message shown, res.json is created>
```

# Handling uncrustify config content

## Steps to reproduce

5. Launch the uncrustify analyzer like follows (use relative path to `test.c` file and add `--report-html` option):

```shell
$ python -m universum.analyzers.uncrustify --files test.c --result-file res.json --cfg-file $(pwd)/uncrustify.cfg --report-html
```

## Expected result

The analyzer finishes successfully or produces actionable error message.

## Actual result

The analyzer produces the following error message:

```
local variable 'wrapcolumn' referenced before assignment
```

# Handling uncrustify config formatting

## Steps to reproduce

6. Edit `uncrustify.cfg` file as follows:
```
code_width 110
input_tab_size 4
```
7. Run the uncrustify analyzer exactly like before:
```shell
$ python -m universum.analyzers.uncrustify --files test.c --result-file res.json --cfg-file $(pwd)/uncrustify.cfg --report-html
```

Note that this uncrustify config is correctly handled by the uncrustify itself.

## Expected result

The analyzer finishes successfully or produces actionable error message.

## Actual result

The analyzer produces the following error message:
```
list index out of range
```

## Details

The analyzer can only handle the following config:
```
code_width = 110
input_tab_size = 4
```

The spaces and equal sign are mandatory.

# Environment

 - OS: Ubuntu 22.04
 - Python: 3.10
 - Universum: 0.19.15, commit id 957b90ee

[...]

PR #771, opened 2023-03-16
fix(uncrustify): usability issues

i-keliukh:
Fixed the following issues in uncrustify analyzer:
- issue with not finding the formatted file if if was passed to the
analyzer by using absolute path
- issue with not initializing variables leading to exception if their
value are not found in uncrustify config file
- issue with not handling all variants of the uncrustify config file
formatting


i-keliukh:
Fixed #770

k-dovgan (approved):

k-dovgan:
Do we want to add tests on this fixes?

i-keliukh:
I think tests are not needed:
 - these issues are found in quite specific part of functionality - uncrustify analyzer - and all of them happen due to interaction with uncrustify
 - the uncrustify part of the interface is not documented well
 - these are usability issues, not bugs

In conclusion, we wouldn't benefit much from these tests.

PR #791, opened 2023-04-24
test(artifact_collector): artifacts preprocessing for conditional steps

miltolstoy:


k-dovgan (approved):

tests/test_conditional_steps.py
```diff
@@ -46,6 +50,10 @@ def check_conditional_step(self, steps_info: StepsInfo) -> None:
         self._check_executed_step_artifacts_present(steps_info)
         self._check_not_executed_step_artifacts_absent(steps_info)
 
+    def check_conditional_step_fail(self, steps_info: StepsInfo) -> None:
```
i-keliukh, line 53: The function name implies that conditional step fails.
However, the conditional step should not fail the execution in any case.
Also, in this specific scenario, the failure happens due to artifact presence before the step executed.
So, technically it fails due to conditional step, but in the conditional step.
I understand that this function is a copy of the `check_conditional_step` function, but the changed name is slightly confusing.

Do we need a separate function at all?

k-dovgan (approved):

PR #800, opened 2023-05-15
fix(test): hotfix GeckoDriver bug with wrong Firefox path

k-dovgan:
See details here: https://github.com/mozilla/geckodriver/issues/2062


k-dovgan:
Not required for proper work

Issue #280, opened 2019-03-11
Incorrect versions of dependencies


i-keliukh:
## Description

The setup.py incorrectly lists version of dependencies:
https://github.com/Samsung/Universum/blob/9c38d4b5329ad596fed9ff00a77d6793e6e98132/setup.py#L42-L62

Specifically, Universum requires the following versions of packages:
- urllib3: 1.13 (default in Ubuntu 14.04: 1.7.1)
- docker-py: 2.5.0 (for testing only)
- pytest 3.3

At the same time, tests pass with pytest==3.7, but in setup.py this version is not allowed.

## Steps to reproduce

1. Install Ubuntu 14.04
2. Install Universum and its development environment (docker, `sudo pip install .[test]`, etc)
3. Run tests with `make test`

## Expected result

Tests succeeded.

## Actual result

Tests fail due to unavailable api in urllib3 and docker-py.

## Environment

- OS: Ubuntu 14.04
- Universum version: 0.17.0
- Urllib3: 1.7.1
- Docker-py: 2.2.1
- requests: 2.11.1
- chardet: 2.0.1
- pytest: 3.0.7

## Additional context

For rechecking the issue it is required to use minimal version of the dependencies. Otherwise, docker-py can in turn upgrade urllib3 to the version, which is newer than required for the Universum itself.

By default, pip refuses to upgrade urllib3, because it is installed with distutils (apt-get). Use `-I` command-line parameter to force pip to still upgrade urllib3:
```
sudo pip install urllib3==1.13 -I
```

i-keliukh:
Please note I'm not sure about the minimum required version of the pytest. This has to be checked separately.

i-keliukh:
For the record: it turned out pytest 3.7.3 breaks our tests.

a-osipov-toxa:
The package `p4python` is successfully installed (`python3.7 -m pip install p4python`) with the following environment:
Ubuntu: 16.04.6
Openssl: 1.0.2g
Python: 3.7
Pip: >=19.0 (this version provides installation without compilation and access to ftp.perforce.com)

i-keliukh:
> Pip: >=19.0 (this version provides installation without compilation and access to ftp.perforce.com)

@a-osipov-toxa, does it mean that with pip 18.x the installation fails?

a-osipov-toxa:
> @a-osipov-toxa, does it mean that with pip 18.x the installation fails?

[...]

PR #802, opened 2023-05-16
fix(test): outdated dependencies

k-dovgan:
Latest selenium version is 4.0.9; latest urllib3 version is 2.0.2. However, used selenium version is 3.141, and latest non-failing urllib3 version for running FF with this selenium version is 1.26.15


miltolstoy (approved):

k-dovgan:
Relates to #280

PR #809, opened 2023-05-29
test(configuration_support): test Step->Configuration update for conditional step branches

miltolstoy:
Postponed tests update for https://github.com/Samsung/Universum/pull/808 change.

### DIFF
--- a/tests/test_conditional_steps.py
+++ b/tests/test_conditional_steps.py
@@ -13,8 +13,8 @@ from .utils import LocalTestEnvironment
 
 class StepsInfo:
     conditional_step: Step
-    true_branch_step: Optional[Step]
-    false_branch_step: Optional[Step]
+    true_branch_steps: Optional[List[Step]]
+    false_branch_steps: Optional[List[Step]]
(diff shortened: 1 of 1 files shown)


tests/test_conditional_steps.py
```diff
@@ -98,47 +95,54 @@ def _build_step_command(files_to_create, exit_code) -> List[str]:
...
             "configs = Configuration([conditional_step])"
         ]
         config: str = "\n".join(config_lines)
 
         self.configs_file.write_text(config, "utf-8")
 
+    @staticmethod
+    def _build_branch_configuration_string(steps: Optional[List[Step]]) -> str:
+        if not steps:
+            return "None"
```
k-dovgan, line 120: Is this string intended to return value that looks alike to actual `None`?

i-keliukh (approved):

tests/test_conditional_steps.py
miltolstoy: Yes, as this value is inserted in a Python config file in the `_write_config_file()` method and then interpreted by Universum.

PR #821, opened 2023-08-25
fix(swarm): quickfix for incorrect '404' response

k-dovgan:
Currently Swarm server returns '404' when adding comments to newly added files in review, causing to report malfunction.

### DIFF
--- a/universum/modules/vcs/swarm.py
+++ b/universum/modules/vcs/swarm.py
@@ -6,6 +6,7 @@ from ..error_state import HasErrorState
 from ..output import HasOutput
 from ..reporter import ReportObserver, Reporter
 from ...lib import utils
+from ...lib.utils import Uninterruptible
 from ...lib.ci_exception import CiException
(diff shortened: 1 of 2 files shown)


universum/modules/vcs/swarm.py
```diff
@@ -193,14 +194,13 @@ def report_start(self, report_text):
...
     def code_report_to_review(self, report):
-        for path, issues in report.items():
-            abs_path = os.path.join(self.client_root, path)
-            if abs_path in self.mappings_dict:
-                for issue in issues:
-                    self.post_comment(issue['message'],
-                                      filename=self.mappings_dict[abs_path],
-                                      line=issue['line'],
-                                      no_notification=True)
+        with Uninterruptible(self.out.log_error) as run:
```
i-keliukh, line 197: What is the benefit of using context manager in comparison to handling one exception?

miltolstoy (commented):
LGTM

PR #835, opened 2023-11-24
feat(github): add postcommit workflow

k-dovgan:


i-keliukh (approved):
Is it possible to reuse the same workflow for both pre- and post-commit check?

k-dovgan:
> Is it possible to reuse the same workflow for both pre- and post-commit check?

Not recommended, as pre-commit uses very specific PR payload to extract values and make comments directly to review.
They are triggered by different events and processed differently, so uniting them in one workflow will only complicate things.

miltolstoy (commented):
LGTM

PR #860, opened 2026-05-14
add(gha): test Python versions and update docs

k-dovgan:
This workflow is added for illustrative purposes, for easy checking different Python versions compatibility with current Universum or potential changes. It can be run manually on demand, and is advised to be checked for any new Universum release.


k-dovgan:
Workflow example: https://github.com/Samsung/Universum/actions/runs/25885374074

i-keliukh:
I see the workflow failed on "Setup Python" step for 3.6 and 3.7 and on making docker images for 3.8 and 3.9. The updated doc says that tests require Python >= 3.12. So it is not clear what is the reason of having 3.6-3.11 tested.
Another issue - the `nox_only` decorator is not updated. I think we should either change it or remove corresponding tests.

k-dovgan:
> I see the workflow failed on "Setup Python" step for 3.6 and 3.7 and on making docker images for 3.8 and 3.9. The updated doc says that tests require Python >= 3.12. So it is not clear what is the reason of having 3.6-3.11 tested.

I think I might add some kind of switch in tests between "under 3.12 - higher" in a different PR. I'd leave it as is now for active Python versions (3.10 and higher), as this WF is more _illustrative,_ than "required": you can see I only left "workflow_dispatch" trigger, so it only can be run manually.

> Another issue - the `nox_only` decorator is not updated. I think we should either change it or remove corresponding tests.

Same here: the decorator is not actually connected to nox, it's about regression. It is now used correctly, and it needs to be renamed — but as I see the tests need to be fixed in several places for Python other than 3.12, I'd leave it to the next PR.

i-keliukh:
> This WF is more illustrative, than "required": you can see I only left "workflow_dispatch" trigger, so it only can be run manually.

No problem. Please describe the purpose of the workflow in the PR description.

i-keliukh (approved):

PR #861, opened 2026-06-10
fix(p4): rare scenario of file version conflict

k-dovgan:
Also refactored test structure a little bit, it was becoming unreadable.
Splat "regression" into just regression and p4-related regression


i-keliukh (approved):

miltolstoy:
LGTM

PR #862, opened 2026-06-15
fix(test): switch to venv in tests

k-dovgan:
- Switch to venv in tests according to recommendations
- Make post-commit optional for non-main branches
- Also fix doctest for Python3.14 new warning format

### DIFF
--- a/tests/test_code_report.py
+++ b/tests/test_code_report.py
@@ -13,7 +13,8 @@ from .utils import python, python_version
 
 @pytest.fixture(name='runner_with_analyzers')
 def fixture_runner_with_analyzers(docker_main: UniversumRunner):
-    docker_main.environment.install_python_module("universum[test]")
+    # can't use "install python module" due to "[test]" part
+    docker_main.environment.assert_successful_execution("pip install -U universum[test]")
     docker_main.environment.assert_successful_execution("apt install -y uncrustify clang-format")
     yield docker_main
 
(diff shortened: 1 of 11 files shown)


tests/test_code_report.py
```diff
@@ -11,11 +11,11 @@
 from .deployment_utils import UniversumRunner
 from .utils import python, python_version
 
-
 @pytest.fixture(name='runner_with_analyzers')
 def fixture_runner_with_analyzers(docker_main: UniversumRunner):
-    docker_main.environment.install_python_module("pylint")
-    docker_main.environment.install_python_module("mypy")
+    # This particular test set needs system-wide Python, because analyzers somehow don't support venv
```
miltolstoy, line 16: As I know, you can just use `pip` and `python` bins directly from the venv dir, and it should work without regular venv activation.
What was the exact issue with the analyzers in your case?

k-dovgan: In this particular case we pass python version to analyzers and it launches like "python{version} -m <analyzer>". So trying to launch this tests launches the system-wide python bin for running analyzer, while using the venv bin for handling the test

k-dovgan: Shortly: pip-in-venv logic works if you are not trying to launch it in docker via python framework :)

miltolstoy (approved):

